Issue #55, opened 2020-08-26
Ignore events (e.g. availability blocks)


jaden-young:
**Is your feature request related to a problem? Please describe.**
Work asks for calendar events for work from home (WFH)/availability. Status bar ends up showing `WFH Now` up until ~15 minutes before an actual event starts.

**Describe the solution you'd like**
I'd like to be able to have some way to prevent an event from showing on the status bar, I want the next event to show, e.g. `Team meeting in 2hrs` even if the ignored event is currently happening. I'd also want `Join next meeting` to skip the ignored event.

**Describe alternatives you've considered**
For WFH/availability blocks e.g. Google calendar has some options like working hours and appointment slots, but alas, thou shalt submit to the HTML email wielding branch of the organization.

**Additional context**
![without dropdown](https://user-images.githubusercontent.com/9858187/91252709-b5e86780-e723-11ea-83fd-67f3782a6083.png)
![with dropdown](https://user-images.githubusercontent.com/9858187/91252729-c26cc000-e723-11ea-9edf-d2d243c44375.png)

leits:
Wow. Great idea, @jaden-young !

Do you have any things cleaning the filter ignoredCalendarItemIDs?
It looks like they will mostly just filling up.
And maybe it makes sense to do it only for recurring events?

Waiting for your feedback. Thanks for the feature!)

jaden-young:
@letis I think your comment was meant for the PR in #56, moving discussion there.

fuzzy76:
Would #84 help?

leits:
@jaden-young, please, take a look on @fuzzy76 question

jaden-young:
I think #84 is a great/elegant solution to this issue! I'd say to close #55 and #56 in favor of #84. The remaining functionality provided by manually ignoring events would just be swept up by declining attendance.

jgoldhammer:
We will close this feature requests as suggested by @jaden-young and will follow #84 . Thanks!

lilyball:
#84 is not enough. I want the status item to always show my next event, even if there's no meeting link (as I find MeetingBar a much better way of being aware of upcoming events than Calendar). However, I still want to filter out specific events.

In particular, when I have personal events during work time, I need to put an event on my work calendar to mark myself as busy. I typically title these events "Blocked out", for the same duration as my personal event, with no other information. And as a result, MeetingBar shows both events in the same time slot. It also can't seem to decide which one to prioritize.

[...]

PR #56, opened 2020-08-26
Add ability to ignore events

jaden-young:
Prevents showing ignored events in status bar title and removes them from eligibility for "join next meeting." Note that this operates on calendar items, not specific events, so all future events for that item are also ignored.

### Status
**IN DEVELOPMENT**

### Description
Quick buggy fix of #55 that works for me personally. Introduces an empty `Status: ` with a debug string for events that aren't `Accepted`. Not sure if this approach is the best, just throwing it out there in case anyone finds it useful.

image
image

### DIFF
--- a/MeetingBar/AppDelegate.swift
+++ b/MeetingBar/AppDelegate.swift
@@ -27,6 +27,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     var disablePastEventObserver: DefaultsObservation?
     var declinedEventsAppereanceObserver: DefaultsObservation?
     var showEventsForPeriodObserver: DefaultsObservation?
+    var ignoredCalendarItemIDsObserver: DefaultsObservation?
+    var ignoredEventIDsObserver: DefaultsObservation?
 
     var preferencesWindow: NSWindow!
 
@@ -129,7 +131,16 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
                 NSLog("Changed showEventsForPeriod from \(change.oldValue) to \(change.newValue)")
                 self.statusBarItem.updateMenu()
             }
-
+            self.ignoredCalendarItemIDsObserver = Defaults.observe(.ignoredCalendarItemIDs) { change in
+                NSLog("Changed ignoredCalendarItemIDs from \(change.oldValue) to \(change.newValue)")
+                self.statusBarItem.updateMenu()
+                self.statusBarItem.updateTitle()
+            }
+            self.ignoredEventIDsObserver = Defaults.observe(.ignoredCalendarItemIDs) { change in
+                NSLog("Changed ignoredEventIDs from \(change.oldValue) to \(change.newValue)")
+                self.statusBarItem.updateMenu()
+                self.statusBarItem.updateTitle()
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


MeetingBar/AppDelegate.swift
```diff
@@ -27,6 +27,8 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     var disablePastEventObserver: DefaultsObservation?
     var declinedEventsAppereanceObserver: DefaultsObservation?
     var showEventsForPeriodObserver: DefaultsObservation?
+    var ignoredCalendarItemIDsObserver: DefaultsObservation?
+    var ignoredEventIDsObserver: DefaultsObservation?
```
jaden-young, line 31: woops, holdover from trek down ignore once/ignore all future events that ended up being undone

leits:
Wow. Great idea, @jaden-young !

Do you have any things cleaning the filter _ignoredCalendarItemIDs_?
It looks like they will mostly just filling up.
And maybe it makes sense to do it only for recurring events?

Waiting for your feedback. Thanks for the feature!)

jaden-young:
Responding to https://github.com/leits/MeetingBar/issues/55#issuecomment-681095558

> Do you have any things cleaning the filter ignoredCalendarItemIDs?
> It looks like they will mostly just filling up.

There isn't anything sweeping for dead ids, but if the event occurs today it'll show up in the status bar with an `Unignore this event` entry that the user can click to remove it. There's certainly an opportunity to put in some cleanup logic but I'm not sure what the desired behavior would be.

From what I can see these would just fill up a plist file somewhere. Not sure what the length of a `calendarItemIdentifier` is, but if a UUID v4 is 36 bytes, round up to 50 for some overhead, that's around 200 distinct events before this hits 1MB. I pity whoever needs to ignore that many independent events.

> And maybe it makes sense to do it only for recurring events?

Possibly, if it was for a one-time event the user would probably just decline it for the same effect.

jaden-young:
@leits Made the ignore option only show for completed events and fixed the debug string showing for participant status, https://github.com/leits/MeetingBar/pull/56/commits/7f609077d547ca4eb5a9c18db40214b6db18553c is in a workable state. Started down trying to cleanup stale ignore IDs but ran into permissions issues, see the message on https://github.com/leits/MeetingBar/pull/56/commits/9a84d8819bd30b5ead36e33ac075b5ecd8541f84. Do you have any suggestions to get that last commit working?

MeetingBar/AppDelegate.swift
```diff
@@ -167,6 +173,20 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
         }
     }
 
+    private func scheduleCleanupIgnoredEvents() {
+        let activity = NSBackgroundActivityScheduler(identifier: "leits.MeetingBar.cleanupignoredevents")
+
+        activity.repeats = true
+        activity.interval = 5
```
jaden-young, line 180: TODO: increase this interval after getting cleanup actually working

jaden-young:
Hey @leits, I think I figured out the issue with permissions. TL;DR it works if we request access to `EKEntityType.reminder` in addition to `EKEntityType.event`.

The goal is to look at `ignoredEventIDs: Set<string>` and remove from that set events that don't exist (have been deleted) on a user's calendar(s). Errors occurred for both `EKEventStore.event(withIdentifier:)` and `EKEventStore.calendarItem(withIdentifier:)` saying permission denied. The error for `calendarItem` kind of made sense as maybe that could also return `reminders`, but it's strange to me that `event` would bomb out since we definitely have permission for events. There's a reverted workaround commit in there showing querying a swath of events before/after today that worked.

I see you're working on refactoring onboarding/initialization on master so I'll wait until that shakes out to integrate further - HEAD ended up not detecting any of my calendars and not prompting for permission on my machine. Feel free to merge/pick/squash/copy-paste whatever is in here into your refactor, I believe I ticked the "allow changes from maintainers" box.

jgoldhammer:
We will close this feature requests as suggested by @jaden-young and will follow #84 . Thanks!

PR #66, opened 2020-09-08
Allows hiding past events

dev10110:
### Status
READY

### Description
Allows hiding events in the past, instead of just adding a checkmark to the side of the event. 
Exposes an user-option to set this field, defaults to false to maintain past behaviour.


Itaybre (commented):
The file `MeetingBar/MeetingBar.xcodeproj/project.xcworkspace/xcuserdata/devanshagrawal.xcuserdatad/UserInterfaceState.xcuserstate` should be ignore and not pushed

dev10110:
@Itaybre ah that seems to have slipped through indeed. I'll close this pr and open a new one. Thanks!

dev10110:
oh, I realised I could just git rm the file. I believe it should be gone now. I've also removed the gitignore file I temporarily added. I hope this doesnt screw with the history too much!

leits:
@dev10110 
What do you think about combining this option with disablePastEvents?

dev10110:
I don't mind! So some sort of drop down where you select either hide, cross out or allow joining?

leits:
I think this feature has been on the review for too long. I'll think about improving UX separately.
Thanks, @dev10110 !

leits (approved):

dev10110:
my apologies for the delay :)

Issue #34, opened 2020-07-29
Add authuser to url


sirupsen:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I am logged into multiple Google Accounts for Meet. Ideally, it'd figure out which one owns an event, and then add `?authuser=$email` to the URL to avoid users having to switch manually. 

**Describe alternatives you've considered**
This seems like the simplest solution.

**Additional context**

dansiemens:
Duplicate of #24

BrandonSmith:
Seems this is available in latest release. Right?

Itaybre:
@BrandonSmith you are right, it was merged for 1.5

@leits please close this issue

leits:
Thanks, @BrandonSmith, @Itaybre. Closed.

PR #69, opened 2020-09-11
Add Google Account Detection

### LINKED ISSUE #24: Event always opens the meeting with the default account not the calendar-specific account
**Describe the bug**
When selecting an event, Google Meet opens with the default account selected, not the account of the calender.

**To Reproduce**
1. Have multiple accounts logged in and calendars added to MeetingBar
2. Select a calendar event that is not the default logged in

**Expected behavior**
Google Meet should open with the account of the calendar event

**Desktop (please complete the following information):**
 - macOS version: [e.g. 10.15.4]

**Additional context**
There is a `authuser` query parameter which selects the account, but not fully sure how you are pulling in the Event details to make sure a suggestion on the correct account name to integer mapping.
e.g. `https://meet.google.com/abc-defg-xyz?authuser=1`

Itaybre:
### Status
**READY**

### Description
Add the capability to detect Google Account (would like to test this on more accounts, works on the 3 accounts I have currently).
When opening a Google Meet link, it ill add `authuser` parameter with the detected account.
Fixes #24 and #34 

1. Have 2 different Google Accounts, and select `A` as default in Google
2. Open a Meeting for account B
3. You will have to change account

### Notes
This solution is Hacky and likely to break with an update, but doesn't use any private API, so it should be fine if you want to release MeetingBar in the macOS Store

### DIFF
--- a/MeetingBar/Constants.swift
+++ b/MeetingBar/Constants.swift
@@ -110,3 +110,7 @@ enum JoinEventNotificationTime: Int, Codable {
     case minuteBefore = 60
     case fiveMinuteBefore = 300
 }
+
+struct GoogleRegex {
+    static let emailAddress = try! NSRegularExpression(pattern: #""mailto:(.+@.+)""#)
+}
--- a/MeetingBar/Helpers.swift
+++ b/MeetingBar/Helpers.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2020 Andrii Leitsius. All rights reserved.
 //
 import Cocoa
+import EventKit
 
 func getMatch(text: String, regex: NSRegularExpression) -> String? {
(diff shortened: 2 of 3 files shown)


leits:
Great idea, @Itaybre!
Apple maintains they API interfaces for a long time and notifies in advance of changes, so everything is fine from this side.
If Google changes the way it stores the source, then this feature will simply stop working. And the application will continue to work as it worked without this feature.

Correct me if I'm wrong.)

Itaybre:
@leits Apple could stop printing this information inside `EKSource` description, but in the worst case scenario, it will stop working, but no crash should happen

leits:
OK. LGTM!

leits (approved):

jogendra:
That was really great addition @Itaybre ✨ 
I was looking for something like this.

PR #73, opened 2020-09-11
Added check for event URL field

insipidlight:
### Status
READY

### Description
I don't think this is optimal, but it's simple. Using URL from EKCalendarItem, adds url.absoluteString to linkFields, which is then converted back again. I prefer to leave Zoom links in this field.

Tested both a Zoom and a Google Meet link in URL field (3rd field from bottom) in macOS 10.15.6 calendar


leits:
Thanks!
I will take this case with URL into account when I'll need to change this code.

leits (approved):

Issue #59, opened 2020-08-28
Open Calendar event in Submenu


dev10110:
**Is your feature request related to a problem? Please describe.**
Feature Request. 
When I click into the details of an event in the meeting bar, I can't click anything, especially links that are embedded in the event details.

**Describe the solution you'd like**
I would like to be able to click somewhere to open my calendar (hopefully to the specific event) so I can see all the details. 

**Describe alternatives you've considered**
Currently, I just open the calendar separately, find the event and open the links.

dev10110:
Addressed by #60, but not fully - calendar app can be opened, but not to a specific event.

leits:
Thanks, @dev10110!
It's a great idea to open a calendar app for the event!

I think the opening event in the calendar is much more valuable than the calendar opening. 
This can be done easily in other ways, as opposed to opening an event in the calendar.

Did you try to use `calshow:` URI scheme? Maybe it will works as you want.

dev10110:
I tried, but I am not familiar with the URI scheme, and couldn't figure out what the right URI should be.
If you point me in the right direction, I'll implement it!

btw, for mac it seems we need to use `ical`, `calshow` is for iOS (?).

leits:
@dev10110 , looks like I found the way:
`let url = URL(string: "ical://ekevent/\(identifier)")`

Try it when you'll have a time.

dev10110:
@leits  thats awesome!! and it works! 

im just (very) new to Swift in general, and can't figure out how to pass the event identifier as an argument from `StatusBarItemController.swift` into `AppDelegate`. If you point me in the right direction I can try it again.

leits:
@dev10110 added in https://github.com/leits/MeetingBar/releases/tag/v2.0.
Check it out!

PR #79, opened 2020-09-16
Open Event in Calendar

dev10110:
### Status
**IN DEVELOPMENT**

### Description
Tries to open the calendar to a specific event. 

In response to #59

### DIFF
--- a/MeetingBar/AppDelegate.swift
+++ b/MeetingBar/AppDelegate.swift
@@ -184,6 +184,29 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
 
         completionHandler()
     }
+  
+  // just opens the calendar
+  @objc func openCalendar(_: Any? = nil) {
+    NSLog("Open Calendar")
+    let url = URL(string: "ical://")
+    NSWorkspace.shared.open([url!],
+    withAppBundleIdentifier:"com.apple.iCal",
+    options: [],
+    additionalEventParamDescriptor: nil,
+    launchIdentifiers: nil)
+  }
+  
+  // try to open to specific event
+  @objc func openCalendarEvent(_: Any? = nil){
+    let identifier = "EF087808-AC5B-4672-A49E-90FC400D4C61" // HARDCODED identifier works!
+    NSLog("Open Calendar EVENT with identifier \(identifier)")
+    let url = URL(string: "ical://ekevent/\(identifier)")
+    NSWorkspace.shared.open([url!],
+    withAppBundleIdentifier:"com.apple.iCal",
+    options: [],
+    additionalEventParamDescriptor: nil,
+    launchIdentifiers: nil)
+  }
 
     @objc func createMeeting(_: Any? = nil) {
         NSLog("Create meeting in \(Defaults[.createMeetingService].rawValue)")
--- a/MeetingBar/StatusBarItemControler.swift
+++ b/MeetingBar/StatusBarItemControler.swift
@@ -102,6 +102,12 @@ class StatusBarItemControler {
             action: #selector(AppDelegate.createMeeting),
             keyEquivalent: "")
         createItem.setShortcut(for: .createMeetingShortcut)
+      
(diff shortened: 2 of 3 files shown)


MeetingBar/AppDelegate.swift
```diff
@@ -184,6 +184,29 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
...
+    let url = URL(string: "ical://")
+    NSWorkspace.shared.open([url!],
+    withAppBundleIdentifier:"com.apple.iCal",
+    options: [],
+    additionalEventParamDescriptor: nil,
+    launchIdentifiers: nil)
+  }
+  
+  // try to open to specific event
+  @objc func openCalendarEvent(_: Any? = nil){
```
dev10110, line 200: I tried a bunch of things but can't get the event identifier passed in as an argument! 

Could you point me in the right direction?

MeetingBar/StatusBarItemControler.swift
```diff
@@ -192,7 +198,13 @@ class StatusBarItemControler {
             let titleItem = eventMenu.addItem(withTitle: eventTitle, action: nil, keyEquivalent: "")
             titleItem.attributedTitle = NSAttributedString(string: eventTitle, attributes: [NSAttributedString.Key.font: NSFont.boldSystemFont(ofSize: 15)])
             eventMenu.addItem(NSMenuItem.separator())
-
+          
+            // Open in Calendar
+          eventMenu.addItem(
+            withTitle: "Open in Calendar",
+            action: #selector(AppDelegate.openCalendarEvent),
```
dev10110, line 205: `event.eventIdentifier` contains the identifier string needed for `ical://ekevent/\(identifier)`

MeetingBar/StatusBarItemControler.swift
```diff
@@ -102,6 +102,12 @@ class StatusBarItemControler {
             action: #selector(AppDelegate.createMeeting),
             keyEquivalent: "")
         createItem.setShortcut(for: .createMeetingShortcut)
+      
+      self.item.menu!.addItem(
```
Itaybre, line 106: It is relevant to add a button to open the Calendar when you could just use Spotlight?

MeetingBar/AppDelegate.swift
Itaybre: I will take a look at this, I have already implemented this [here](https://github.com/Itaybre/Siguiente/blob/a5479abf9bdacc5cf3b676af1cc4d86b0c7ca425/Siguiente/Helpers/StatusBar/StatusBarManager%2BMenuTarget.swift#L23), so it shouldn't be much different

MeetingBar/AppDelegate.swift
```diff
@@ -184,6 +184,29 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
 
         completionHandler()
     }
+  
+  // just opens the calendar
+  @objc func openCalendar(_: Any? = nil) {
```
Itaybre, line 189: I would prefer not to add more responsibilities to the AppDelegate, what do you think @leits ?
Should we have another class in charge of all this actions?

MeetingBar/AppDelegate.swift
dev10110: Yea I'm a bit mixed on this too, but often when I'm looking at the details for an event I just want to click so the calendar comes up to that event. We can take the option out of the main menu but leave it in the submenu for each event? 

Or if possible make it such that clicking anywhere in the submenu will open the calendar event? But I couldn't figure out how to do that.

Itaybre: You can use the modifier [flags](https://stackoverflow.com/questions/35802811/swift-nsmenuitem-detect-shift-click) to do a different action when clicking with specific keys

leits:
Added in https://github.com/leits/MeetingBar/releases/tag/v2.0.

PR #89, opened 2020-10-04
Bookmark

salling:
### Status
**IN DEVELOPMENT**

### Description
In my team, we like to use one meeting room for quick gettogethers. While it's certainly possible to use the existing MeetingBar functionality for creating a brand new meeting each time, this necessitates the sharing of links etc. With a "bookmark", I can just tell my colleagues to "meet in <xyz>".

I understand @leits may have other thoughts/ideas re: this type of feature, and my implementation is certainly very simplistic. One could imagine supporting multiple bookmarks, etc.

Just scratching my own itch here, really. Would love to see something like this in the app.

1. Configure your bookmark in Preferences
Screenshot 2020-10-04 at 17 37 40
2. A new menu item now becomes available in the shortcut menu:
Screenshot 2020-10-04 at 17 36 43


leits:
It's a great idea! Thanks, @salling. 
As you remarked, It would be good to develop to the many bookmarks concept. 
Would you like to continue developing this feature?

salling:
> It's a great idea! Thanks, @salling.
> As you remarked, It would be good to develop to the many bookmarks concept.
> Would you like to continue developing this feature?

Apologies for the late reply @leits. I do not have the bandwidth to build the multi-bookmark concept.

If someone else is willing to do it now, they should go ahead. I wouldn't mind at all.

If not, I propose tweaking this PR such that the feature can be *extended* in the future with multi-bookmarks. In my opinion, what'd make it easier to pick up and add support for multiple bookmarks is more robust persistence data structure. The way it is, the lucky fellow building multi-bookmark support would have to migrate the old bookmark user defaults into some other structure.

Perhaps you would like to propose some bookmark data structure that you feel would fit your future vision for this feature?

leits (approved):

jgoldhammer:
we will deliver the first iteration of bookmarks (up to 5) in the next version 🎉

PR #91, opened 2020-10-09
Ability to pause the Apple Music app when joining a meeting

DDRBoxman:
### Status
**Ready**

### Description
This is a quick attempt at using scripting bridge to pause the Apple Music app when a meeting is started.
Ideally we can add other music apps fairly easily in a similar way.

To use scripting bridge you have to remove the app sandbox, that's less than ideal.

Honestly mostly threw this together for my own use and was interested in hearing feedback.

Turn on the pause music preference and when launching a meeting the apple music app will pause


leits:
Possibility to run custom AppleScript added in https://github.com/leits/MeetingBar/releases/tag/v2.0
@DDRBoxman check it out!

Issue #103, opened 2020-12-03
Feature Request: (Option to) Hide meetings with no other participants


0bmxa:
**Is your feature request related to a problem? Please describe.**
I have a bunch of meetings in my calendar, that are just blockers or tasks for myself, which are not "real meetings". They all show up in the list, and are a bit distracting from the actual meetings I have with others.

**Describe the solution you'd like**
Hide meetings that have no other participants from the list.

Ideally not by default, but with an option in Preferences to turn this on/off.
(Probably in the `General` under `Show events for [today/tomorrow]`.)

**Describe alternatives you've considered**
Alternative display:
* Instead of hiding them, they could also be grayed out.

Alternative decision which to display:
* Hide meetings with no video call
* Hide meetings with a certain prefix in the title
* (Hide meetings with no other guests)

**Additional context**
N/A, sorry

0bmxa:
I'm also happy to implement this feature myself, but I first wanted to check if there's any interest in this by anyone else, or if anything could be done differently!

jgoldhammer:
This is merged now 🎉

0bmxa:
I'll close this as its merged now!

Issue #96, opened 2020-11-10
Feature Request: Use a different style to display past meetings


tmchow:
**Is your feature request related to a problem? Please describe.**
It can be hard to scan the list of past and upcoming meetings due to the use of the checkmark since the text is all the same style.

**Describe the solution you'd like**
Instead of a checkmark indicating a past meeting, I think it would work better to dim the row like how the "Next Meeting" product works.

jgoldhammer:
Hi Trevin, 

something like this?

![image](https://user-images.githubusercontent.com/3872101/103648168-b52ad200-4f5c-11eb-85ce-bd59c5f5d48b.png)

We have removed the checkmark...

tmchow:
@jgoldhammer yes! that would help immensely by making the list easier to scan for current+future mtgs vs past mtgs.

jgoldhammer:
Thanks for your answer. We will release this as feature for the upcoming 3.0 release. 🎉

leits:
@tmchow,  released! Check it out!

PR #110, opened 2020-12-16
Add option to hide meetings with no other participants

0bmxa:
### Status
**READY**

### Description
Implements #103 "option to hide meetings with no other participants", by introducing a `personalEventsAppereance` setting in `Defaults` and the UI preferences. This applies to all events where `hasAttendees` is false. It offers the same three options as the setting for past events (active/inactive/hide).

**Notes/things to review:**
- The defaults key reuses the `PastEventsAppereance` enum, because I used the same three options. I didn't want to create duplicated code, but now the name doesn't fully match anymore.
- The filtering for the status bar title and user notifications happens in `EventStore`'s `getNextEvent`. Not sure if that's the right place, but it seemed to be the only commonly used place by both.
- Adding a row to `Appearance` in the UI required increasing the window height. This window height could maybe be made dynamic and be based on content height instead (in the future).
- The name "personal events" for events without other attendees is maybe not ideal, but I didn't have any other idea. "Private event" seems to be used by calendar services for the event's visibility, "Lonely event" sounds sad and unproductive, "Single event" like it's the only event on a day… happy about input here! ;)

#### Other changes
- This fixes the displaying of past events when set to "inactive", but the option "show submenu" is turned on. Previously, past events were only correcly shown as "inactive" when the sub menu was disabled, because the sub menu was otherwise not accessible on a disable item. I changed it to use the system's `disabledControlTextColor` for the font, instead of disabling them. This makes the item look like the same, but keeps the sub menu usable.
- I also sorted the files in Xcode's project navigator a little bit by their importance. If this is unwanted, I'm happy to undo that.

Prep
1. Create calendar event A, and invite someone to it
1. Create calendar event B, where you invite nobody

Test
1. Make sure both events show up as usual in the main menu
1. Set the setting for `Events without guests` to `hide` or `show as inactive`.
1. Make sure event B is not showing or showing with muted/gray text
1. Wait until event A starts, make sure you get the notification and the status bar title is set
1. Wait until event B happens, make sure you don't get the notification, and the status bar title ignores event B.

### DIFF
--- a/MeetingBar.xcodeproj/project.pbxproj
+++ b/MeetingBar.xcodeproj/project.pbxproj
@@ -125,19 +125,19 @@
 		144C017B2462D0C3000C9FFC /* MeetingBar */ = {
 			isa = PBXGroup;
 			children = (
+				144C017C2462D0C3000C9FFC /* AppDelegate.swift */,
 				14E4110D246F2F9200F73ACF /* PreferencesView.swift */,
 				148E47E724F400C400399929 /* OnboardingView.swift */,
-				144C017C2462D0C3000C9FFC /* AppDelegate.swift */,
-				144C01802462D0C5000C9FFC /* Assets.xcassets */,
-				140D843B2493B7A50055E1CE /* Extensions */,
 				140D84312493A3FE0055E1CE /* StatusBarItemControler.swift */,
-				144C01852462D0C5000C9FFC /* Main.storyboard */,
-				144C01882462D0C5000C9FFC /* Info.plist */,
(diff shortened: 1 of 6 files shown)


0bmxa:
I just noticed the fix I mentioned in "Other changes" has been requested in #96.

@leits Let me know if you need any help with reviewing/understanding my proposed changes! 🙂

leits (approved):

leits:
Thanks, @0bmxa!

PR #115, opened 2021-01-04
Truncate whitespace

AbeJellinek:
Previously, if you had a meeting called "ABC DEF GHI" and truncated it to four characters long, it would appear as `ABC ...`. Now it shows as `ABC...`, which is the expected behavior.

This commit also fixes a bug that caused names to be truncated to one character longer than the length slider's actual position (`String.index`'s `offsetBy` parameter is funky).

### Status
**READY**

### Description
It's common to have long meeting names on your calendar, and many users with smaller screens will want to use the truncation setting. It works pretty well, but it looks a bit weird when there's whitespace or other punctuation before the ellipsis. This PR updates the truncation behavior to be more consistent with what users expect to see.

Create a calendar event with a long name and try dragging the length slider in MeetingBar's preferences. Make sure the observed length matches up with the slider's position and that whitespace is truncated as expected.

### DIFF
--- a/MeetingBar/Constants.swift
+++ b/MeetingBar/Constants.swift
@@ -13,6 +13,11 @@ struct TitleLengthLimits {
     static let max = 55.0
 }
 
+struct TitleTruncationRules {
+    static let excludeAtEnds = CharacterSet.whitespacesAndNewlines
+            .union(CharacterSet.punctuationCharacters)
+}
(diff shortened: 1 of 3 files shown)


leits:
Thanks, @AbeJellinek!

leits (approved):

PR #137, opened 2021-01-06
feat(create-meeting): custom url for new meetings

jgoldhammer:
### Status
**READY**

### Description
The PR adds the possibility for the user to use a custom url for new meetings.
The link is opened in the standard browser
If the link does not contain a http or https at the beginning, we will add the prefix.
If the link is empty or not valid, an alert is shown to the user.

Go to preferences, services and choose custom url as option for create meeting.
Add a valid url and try to create the meeting via menubar or shortcut.
The link should be opened in the standard browser.

Delete the url from the preference and try to create the meeting.
An alert should be displayed with a hint.

Enter a not valid url in the preference and try to create a meeting.
An alert should be displayed with a hint.


jgoldhammer:
We should discuss if we should use a alert or we should use the notification. The problem with notifications is that the user does not allow the notification, so no notification is displayed in this case. Otherwise the notifications looks a little bit nicer.
What do you think, @leits ?

leits (approved):

MeetingBar/AppDelegate.swift
```diff
@@ -337,6 +337,23 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
...
             openMeetingURL(nil, CreateMeetingLinks.outlook_live)
+        case .url:
+            var url: String = Defaults[.createMeetingServiceUrl]
+
+            if !url.isEmpty, let checkedUrl = NSURL(string: url) {
+                if !url.starts(with: "http://") && !url.starts(with: "https://") {
+                    url = "https://" + url
+                }
+
+                openMeetingURL(nil, URL(string: url)!)
```
leits, line 348: createMeetingServiceUrl can be a URL of services we recognize and open in the native app or Chrome (Meet).
I think we can refactor the getMeetingLink function and use the search here and in bookmarks in the future.

leits:
> We should discuss if we should use a alert or we should use the notification. The problem with notifications is that the user does not allow the notification, so no notification is displayed in this case. Otherwise the notifications looks a little bit nicer.
> What do you think, @leits ?

We send notifications when a user tries to open an event to which a link was not recognized.

I understand the meaning of alerts. But I think our communication should be consistent. And excluding notifications is a user decision that we must respect. If it's a DnD mode, then they will still be in the notification center.

MeetingBar/AppDelegate.swift
jgoldhammer: Can you add more details? I have not fully understand it yet...

jgoldhammer:
> > We should discuss if we should use a alert or we should use the notification. The problem with notifications is that the user does not allow the notification, so no notification is displayed in this case. Otherwise the notifications looks a little bit nicer.
> > What do you think, @leits ?
> 
> We send notifications when a user tries to open an event to which a link was not recognized.
> 
> I understand the meaning of alerts. But I think our communication should be consistent. And excluding notifications is a user decision that we must respect. If it's a DnD mode, then they will still be in the notification center.

Yes, I fully understand. If an userbased action is not working, we have to display a hint to the user that something is not working. So I would vote for an dismissable alert.

MeetingBar/AppDelegate.swift
leits: I mean we can recognize the meeting service instead of pass `nil` to `openMeetingURL` and open the link in the correct application.

We can use `getMeetingLink` function with some refactoring to untie it from the event.

leits:
Do you mean NSAlert or alert style for notification?

jgoldhammer:
> Do you mean NSAlert or alert style for notification?

I would say NSAlert.  A notification can be dismissed- the user must see that there is something wrong configured when he actively clicks something.  Async notifications should stay the macos notifications IMHO.

Alternative 1: Check the link on entering in the preferences - this situation for alerting will never happen?!
Alternative 2:  Maybe there are more smart notification libraries for macos.

leits:
It makes sense.
What do you think about trying to join an event without a recognized meeting link?

About the alternatives. The first looks like a crutch and the second is out of the scope of this feature.

jgoldhammer:
> It makes sense.
> What do you think about trying to join an event without a recognized meeting link?
> 
> About the alternatives. The first looks like a crutch and the second is out of the scope of this feature.

Trying to join the next event is the same for me- the user actively wants to join, but there is no meeting link. If he has not activated the notifications, he will never see the error message.

leits:
Hmm. Alternative 3 looks reasonable.
In your opinion, is the original implementation or alternative 3 the best option? Or do you think we need to keep looking for a solution?

jgoldhammer:
> Hmm. Alternative 3 looks reasonable.
> In your opinion, is the original implementation or alternative 3 the best option? Or do you think we need to keep looking for a solution?

I have implementation alternative 3. NsAlert will be only used when the user has switched off notifications...

I have simplified the url handling for the create new meeting feature and we will allow all kind of urls. The user must enter a valid url with a scheme, e.g. https

MeetingBar/AppDelegate.swift
leits: @jgoldhammer, what about this?

jgoldhammer: I think the links for creating a meeting can be totally  different to the meeting links, so we will not recognize it correctly. By using a custom url the user could also use an url with a registered scheme, e.g. for teams to open the local app. I think, this is only for advanced users, so I am not sure if we should put an open magic between the users and the configured url. WDYT?

leits: Make sense

leits (approved):

leits:
@jgoldhammer, please, resolve the conflicts

leits (approved):

leits (approved):

PR #139, opened 2021-01-07
feat(meeting-service): support zoom native links

jgoldhammer:
### Status
**READY**

### Description
- zoom native links with scheme zoommtg scheme are now recognized.
- the url will be passed directly to be opened- the url will not be modified.
- the zoom native service will get the same icon in the toolbar

Include a sample zoommtg-link, e.g. zoommtg://zoom.us/join?confno=92333341349&uname=Person&pwd=123456 into a calendar event.
The event is recognized as zoom meeting

image

### DIFF
--- a/MeetingBar/AppDelegate.swift
+++ b/MeetingBar/AppDelegate.swift
@@ -384,7 +384,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
             openEvent(nextEvent)
         } else {
             NSLog("No next event")
-            sendNotification("There are no next meetings today", "Woohoo! It's time to make cocoa")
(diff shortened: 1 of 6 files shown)


MeetingBar/AppDelegate.swift
```diff
@@ -384,7 +384,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
             openEvent(nextEvent)
         } else {
             NSLog("No next event")
-            sendNotification("There are no next meetings today", "Woohoo! It's time to make cocoa")
+            sendNotification(title: "There are no next meetings today", text: "Woohoo! It's time to make cocoa", subtitle: "")
```
leits, line 387: Do we need a blank subtitle here?

jgoldhammer: Fixed.

leits (approved):

PR #144, opened 2021-01-11
Remove html tags from meeting description

0bmxa:
### Status
**READY**

### Description
When you use the event description's formatting in Google Meet, it adds HTML tags to the events description/notes, which so far just shows the plain HTML code in MeetingBar. This PR strips those tags.

#### Before
Screen Shot 2021-01-11 at 22 22 48

#### After
Screen Shot 2021-01-11 at 22 23 20

1. Create a meeting in Google Meet.
2. Add a description with a some formatting (e.g. bullet point list, like in the screenshot).
3. Check the meetings' details view in MeetingBar.

### Note
This PR also removes the hardcoded dev team from the project file. This is required to make the `XCConfig/DevTeamOverride.xcconfig` work (great idea btw!), because otherwise the project file always wins. I hope this doesn't break anything for you or on CI.


0bmxa:
@leits I'm very sorry, I didn't pay attention to the PR title. While the old title was technically not wrong (see note), this was not the actual ("visible") change from this PR. :D

If you want me to split this into two PRs, let me know!

MeetingBar/Extensions/String.swift
```diff
@@ -48,4 +48,16 @@ extension String {
         }
         return self
     }
+
+    /// Returns a version of the string with all HTML tags removed, if any.
+    /// - Returns: The string without HTML tags.
+    func htmlTagsStripped() -> String {
+        let data = self.data(using: .utf8)!
+        let attributedSelf = NSAttributedString(
```
jgoldhammer, line 56: Looks good to me! Should we wrap the conversion into a try-catch? 
https://www.hackingwithswift.com/example-code/system/how-to-convert-html-to-an-nsattributedstring

MeetingBar/Extensions/String.swift
```diff
@@ -48,4 +48,16 @@ extension String {
         }
         return self
     }
+
+    /// Returns a version of the string with all HTML tags removed, if any.
+    /// - Returns: The string without HTML tags.
+    func htmlTagsStripped() -> String {
```
jgoldhammer, line 54: Technically it is not a stripping process - it is more a conversion, right?

MeetingBar/Extensions/String.swift
0bmxa: That's interesting. That `init` they mention there seems to be *almost* the same, but throwing:
https://developer.apple.com/documentation/foundation/nsattributedstring/1524613-init
The one above returns an optional instead:
https://developer.apple.com/documentation/foundation/nsattributedstring/1535412-init

You think the other one has any benefit? I can replace them if you want.

In any case I absolutely agree this should be wrapped in a condition, even if we keep the optional one, and not be force-unwrapped!

MeetingBar/Extensions/String.swift
0bmxa: Hmm, I thought about that as well.

It of course foremost takes a `String` with HTML tags and returns a (non-attributed) `String` without any of them. So 98% of HTML tags are just plain stripped.

However, it preserves the whitespace, because otherwise it'd become unreadable (the screenshot above would become `The Agenda for today will be:ArriveSocializeTalk businessEatGood Night`). That's also why I went the more expensive way via `NSAttributedString` than just a search-and-replace.

I feel a name like `htmlTagsConverted()` or so implies that more HTML features (bold, italic, color, size, …) are actually preserved, than just whitespace, and might potentially be more misleading to other contributors.

But if you're afraid it's not clear enough we could change it maybe to something like
`htmlTagsStripped(preservingWhitespace: Bool = true)`
What do you think of that?

MeetingBar/Extensions/String.swift
jgoldhammer: Thats both fine for me. Just let us have a fallback with the original description if the stripping/conversion does not work.

MeetingBar/Extensions/String.swift
jgoldhammer: yes, from that point of view you are right. lets have it this way like you have implemented. Thanks for the explanation!

MeetingBar/Extensions/String.swift
```diff
@@ -48,4 +48,17 @@ extension String {
...
+        if let data = self.data(using: .utf8), let attributedSelf = NSAttributedString(
+            html: data,
+            options: [.documentType: NSAttributedString.DocumentType.html],
+            documentAttributes: nil
+        ) {
+            return attributedSelf.string
+        }
```
0bmxa, line 61: @jgoldhammer updated!
any opinion about the indentation?

jgoldhammer: For me thats fine now. I will test locally and come back with a merge...

jgoldhammer:
Hmm, I have tested with a local calendar with macos and an event with a bullet list:

The event looks like this in macos calendar
image

and ends up in 
image

I have tested with M365 calendar too and html stripping breaks the formatting.

Tested with Google Calendar and Google Meet and planing in google calendar- I totally see your point. The html formatting breaks...

image

Even in macos Calendar this is wrong:
image

What should we do here? Try to fix only google calendar event entries? WDYT @leits ?

jgoldhammer:
We could check for event.calendar.source.title and match for "Google" or we check if the notes contain any known html element and strip only if it is containing html!? WDYT?

0bmxa:
@jgoldhammer good catch, i didn't test for that!

I just checked, and sadly filtering for Google doesn't fully solve it, because Google only starts using HTML tags when you using formatting. For notes with line breaks only they don't use HTML. See screenshots here:

| Google Calendar | Apple Calendar |
|---|---|
|  |  |

I guess the most "correct" way is first testing whether the string actually contains any HTML, and not applying the `NSAttributedString` if not, maybe like so:
```Swift
let containsHTMLTags = self.range(of: "<.*>", options: .regularExpression) != nil
```
Not sure that's the best way to test for HTML, though, as it technically only tests for the existence of `<` and `>` with something in between, but probably the best we can get.

At least it'd then also work with all kinds of calendar sources.

What do you think?

jgoldhammer:
> @jgoldhammer good catch, i didn't test for that!
> 
> I just checked, and sadly filtering for Google doesn't fully solve it, because Google only starts using HTML tags when you using formatting. For notes with line breaks only they don't use HTML. See screenshots here:
> 
> Google Calendar	Apple Calendar
> 	
> I guess the most "correct" way is first testing whether the string actually contains any HTML, and not applying the `NSAttributedString` if not, maybe like so:
> 
> ```swift
> let containsHTMLTags = self.range(of: "<.*>", options: .regularExpression) != nil
> ```
> 
> Not sure that's the best way to test for HTML, though, as it technically only tests for the existence of `<` and `>` with something in between, but probably the best we can get.
> 
> At least it'd then also work with all kinds of calendar sources.
> 
> What do you think?

Thanks for the Iivestigation. Sounds reasonable!
Maybe something like

```
private func isHtml(_ value:String) -> Bool {

        let htmlTest = NSPredicate(format:"SELF MATCHES %@", "<[a-z][\\s\\S]*>")
        return htmlTest.evaluate(with: value)
    }
```

jgoldhammer:
> > @jgoldhammer good catch, i didn't test for that!
> > I just checked, and sadly filtering for Google doesn't fully solve it, because Google only starts using HTML tags when you using formatting. For notes with line breaks only they don't use HTML. See screenshots here:
> > Google Calendar	Apple Calendar
> > 	
> > I guess the most "correct" way is first testing whether the string actually contains any HTML, and not applying the `NSAttributedString` if not, maybe like so:
> > ```swift
> > let containsHTMLTags = self.range(of: "<.*>", options: .regularExpression) != nil
> > ```
> > 
> > 
> > Not sure that's the best way to test for HTML, though, as it technically only tests for the existence of `<` and `>` with something in between, but probably the best we can get.
> > At least it'd then also work with all kinds of calendar sources.
> > What do you think?
> 
> Thanks for the Iivestigation. Sounds reasonable!
> Maybe something like
> 
> ```
> private func isHtml(_ value:String) -> Bool {
> 
>         let htmlTest = NSPredicate(format:"SELF MATCHES %@", "<[a-z][\\s\\S]*>")
>         return htmlTest.evaluate(with: value)
>     }
> ```

Do you want to change it this way and update the PR, @0bmxa?

0bmxa:
> Do you want to change it this way and update the PR, @0bmxa?

yes, totally! will do in a minute.
sorry for the delay, today was busy ;)

0bmxa:
> Maybe something like
> 
> ```
> private func isHtml(_ value:String) -> Bool {
> 
>         let htmlTest = NSPredicate(format:"SELF MATCHES %@", "<[a-z][\\s\\S]*>")
>         return htmlTest.evaluate(with: value)
>     }
> ```

@jgoldhammer updated!
i went with a computed property and replaced `\s` with just tab and space, so it doesn't match newlines (not sure if NSPredicate has `/m` set…?)

jgoldhammer:
> > Maybe something like
> > ```
> > private func isHtml(_ value:String) -> Bool {
> > 
> >         let htmlTest = NSPredicate(format:"SELF MATCHES %@", "<[a-z][\\s\\S]*>")
> >         return htmlTest.evaluate(with: value)
> >     }
> > ```
> 
> @jgoldhammer updated!
> i went with a computed property and replaced `\s` with just tab and space, so it doesn't match newlines (not sure if NSPredicate has `/m` set…?). hope that's alright!

Nice- works for me with my testdata. So I am going to merge this!
Thanks for the contribution @0bmxa ! This will help others, too and we are better than the macos calendar now 🎉

jgoldhammer:
@leits can you please review and merge if this is fine for you?

leits:
Great work, @0bmxa, @jgoldhammer! Thanks!

leits (approved):

PR #152, opened 2021-01-17
Feat/150-pending events

jgoldhammer:
### Status
**READY**

### Description
The behaviour for events with pending status (the user has not answered the invitation yet) can be controlled now. There is a new user option in the apperance dialog.

There are 4 options now:

- show - the normal behaviour - event is shown like any other event- notifications are sent
- show underlined - the event is shown underlined in the statusbar and in the menu- notifications are sent 
- show inactive - the event is hidden in the statusbar and shown inactive in the menu - no notifications are sent 
- hide - the event will be completely hidden in the statusbar and menu - no notifications are sent

This can be tested by creating a new meeting with another account and invite you via your mailadress.


jgoldhammer:
@leits  - I am still not sure about the following things:
- the UX in the preference dialog - do you have any idea how to layout the stuff better
- the option show as inactive - what do you think about the current behaviour? Should we hide it from statusbar and sent no notification, but the event is still shown in the menu as inactive. I think, this would align more with the other stuff, right?

MeetingBar/Constants.swift
```diff
@@ -167,6 +167,13 @@ enum AlldayEventsAppereance: String, Codable, CaseIterable {
     case hide
 }
 
+enum PendingEventsAppereance: String, Codable, CaseIterable {
+    case show
+    case show_inactive
+    case show_underlined
```
leits, line 173: What happens if we already have an underline with time?

I think this is too complicated an option and we can do without it.

MeetingBar/Extensions/EventStore.swift
```diff
@@ -39,18 +39,29 @@ extension EKEventStore {
 
 
         for calendarEvent in calendarEvents {
+            var addEvent = false
+
             if calendarEvent.isAllDay {
                 if Defaults[.allDayEvents] == AlldayEventsAppereance.show {
                     filteredCalendarEvents.append(calendarEvent)
```
leits, line 46: It needs to be replaced here as well.

MeetingBar/Constants.swift
jgoldhammer: Underlined means that the text has an underline style with dots. But I can skip it because it is not clear enough for the user.

image

MeetingBar/Extensions/EventStore.swift
jgoldhammer: Ok, I will filter out with option hidden completely. If the user selects "show as inactive" we will display it as menu entry, but it will never be shown in the statusbar. Ok?

MeetingBar/Constants.swift
leits: Got it, sorry!

MeetingBar/Extensions/EventStore.swift
leits: Ok

jgoldhammer:
By the way, I have adapted the preference view and rearrange the appreance tab a little bit.

image

WDYT @leits ?

MeetingBar/Extensions/EventStore.swift
jgoldhammer: This is solved now.

leits: Originaly, I mean that `filteredCalendarEvents.append(calendarEvent)` in this line we need to replaced to `                      addEvent = true`.
But your last change, not related to this replacement, is also correct!

leits:
New order looks better!

MeetingBar/Extensions/EventStore.swift
jgoldhammer: fixed.

leits (approved):

Issue #104, opened 2020-12-07
Bug: App Icon is too big compared to other Big Sur Icons


melgu:
**App version**
2.0.5

**Describe the bug**
The Icon is much bigger than the standard Big Sur icon size and therefore seems a bit out-of-place.

**To Reproduce**
1. Download the app

**Expected behavior**
The icon has the same size as other icons on Big Sur

**Screenshots**
<img width="568" alt="Bildschirmfoto 2020-12-07 um 09 36 27" src="https://user-images.githubusercontent.com/9601403/101328356-2ef17200-3870-11eb-91fa-1cbc53a8b960.png">

**Desktop (please complete the following information):**
 - macOS version: 11.0.1

jgoldhammer:
What do you think @leits? Do you have the possibility to check that? I am still on Catalina on both macbooks...

leits:
Me too, @jgoldhammer. But I have an icon with margins from the author. As soon as I have a chance, I'll try to fix it.

leits:
@melgu, released! Check it out!

melgu:
Looks good! Love the other 3.0 changes as well 😃

PR #155, opened 2021-01-19
Correctly resized and reshaped app icons for Big Sur

JackHinkle:
Downloaded Apple's official Big Sur app icon template from https://developer.apple.com/design/resources/. Used Sketch to place the existing icon and export the correctly sized and shaped assets. Replaced existing assets.


leits:
Thanks, @JackHinkle!
Ref: #104

PR #157, opened 2021-01-21
feat(bookmarks): allow move and ask on delete

jgoldhammer:
### Status
**READY**

### Description
- users can now arrange the order of the bookmarks by drag and drop in the list
- users getting asked if they want to delete a bookmark

#### Rearrange
- Create some bookmarks
- Rearrange by holding mouse button on one bookmark and move up or down
- Test the new order in the menu

#### Delete
- create a bookmark
- try to delete the bookmark
- the user will get a dialog to confirm deletion


leits (approved):

leits:
Awesome!

PR #169, opened 2021-02-05
Move browser link opening into URL extension

0bmxa:
### Status
**READY**

### Description
Hello @leits @jgoldhammer!
Not the most important update, but I felt there's a bit much in the global namespace, so I started moving some helper functions out of there. 

Specifically:
- Added an URL extension, with `openIn(browser:)` and `openInDefaultBrowser()`
- Replaced all `openLinkIn` helpers to use those two
- Renamed `ChromeExecutable` to `Browser` (does that mess with Defaults?)
- Added url property to `Browser`
- Made the list in `ServicesTab` dynamic

If you like this (and if I find time) I'd potentially look into some more options to clean up the global scope a bit. Let me know what you think!

- Test that link opening still works
- Test that the Meet browser from Defaults still works
- Look at the code and let me know if you think that's a good idea and the direction you want to go in!


leits (approved):

leits:
Great work, @0bmxa!
It's not a user-facing feature but it's important for project development. Thanks!

Issue #171, opened 2021-02-08
Bug: Issues with service icons (Big Sur issue?)


nevadascout:
App version 3.0.0.
macOS version: Big Sur 11.1

When I go to the dropdown menu in the app settings to change the icon this is what happens:

(Interestingly, all of the icons display correctly in the menu bar)

https://user-images.githubusercontent.com/377089/107212668-95794480-6a07-11eb-87fc-6cd898414c73.mov

leits:
Thanks for the report, @nevadascout!
I'll fix this soon!
@jgoldhammer, jfyi

jgoldhammer:
Maybe only the size for the icon is missing?

```
...
Image(nsImage: getImage(EventTitleIconFormat.eventtype.rawValue)).resizable()
                                .frame(width: 16.0, height: 16.0)
...
    func getImage(iconName){
        let icon = NSImage(named: iconName)
        icon!.size = NSSize(width: 16, height: 16);
        return icon
    }

```

Maybe the frame with width and height is not enough. WDYT?

leits:
Fix checked by @yarsanich on Big Sure:
![зображення](https://user-images.githubusercontent.com/12017826/107423491-095d3f00-6b25-11eb-9556-754a278a1e3e.png)
Works perfectly!

PR #174, opened 2021-02-08
fix(icons): try to fix the unconstrained size in big sur #171

jgoldhammer:
### Status
**READY**

### Description
This PR should fix hopefully the issue 171. 

- Test with Big Sur
- Navigate to the preference dialog and open Apperance tab
- Click on the dropdown icon in the statusbar section

Due to usage of Catalina I cannot test it, but in other areas we also have set a size, so this will probably fix it.

### DIFF
--- a/MeetingBar/Views/Preferences/AppearanceTab.swift
+++ b/MeetingBar/Views/Preferences/AppearanceTab.swift
@@ -37,25 +37,25 @@ struct StatusBarSection: View {
             HStack {
                 Picker("Icon", selection: $eventTitleIconFormat) {
                     HStack {
-                        Image(nsImage: NSImage(named: EventTitleIconFormat.calendar.rawValue)!).resizable()
+                        Image(nsImage: getImage(iconName: EventTitleIconFormat.calendar.rawValue)).resizable()
                             .frame(width: 16.0, height: 16.0)
                         Text("\u{00A0}Calendar icon")
                     }.tag(EventTitleIconFormat.calendar)
 
                     HStack {
-                        Image(nsImage: NSImage(named: EventTitleIconFormat.appicon.rawValue)!).resizable()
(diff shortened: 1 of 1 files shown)


MeetingBar/Views/Preferences/AppearanceTab.swift
```diff
@@ -37,25 +37,25 @@ struct StatusBarSection: View {
...
+                        Image(nsImage: getImage(iconName: EventTitleIconFormat.appicon.rawValue)).resizable()
+                            .frame(width: 16.0, height: 16.0)
                         Text("\u{00A0}App icon")
                     }.tag(EventTitleIconFormat.appicon)
 
                     HStack {
-                        Image(nsImage: NSImage(named: EventTitleIconFormat.eventtype.rawValue)!).resizable()
-                                .frame(width: 16.0, height: 16.0)
+                        Image(nsImage: getImage(iconName: EventTitleIconFormat.eventtype.rawValue)).resizable()
+                            .frame(width: 16.0, height: 16.0)
```
leits, line 53: Do we need this `frame` if we modify NSImage size directly?

leits:
Thank you for finding time to fix this, @jgoldhammer!

MeetingBar/Views/Preferences/AppearanceTab.swift
jgoldhammer: I do not know really. Maybe it is the wrong way and we should do this?

`Use a Frame and aspectRatio set to .fit`
https://blckbirds.com/post/swiftui-how-to-frame-an-image/

Maybe we have to distribute test builds to let others test with Big Sur ??

leits: Ok. I'll find someone with Big Sure 👌

leits (approved):

Issue #176, opened 2021-02-09
Feature: Option to only show events from the near future


igorkulman:
**Is your feature request related to a problem? Please describe.**
If I have a say sync call at 4pm I do not need to see it in the morning, it only becomes relevant for me to see it at about 3pm. In the morning it just makes me nervous seeing it :)

**Describe the solution you'd like**
Currently there is an option in the app to show events for "Today" or "Today and tomorrow". I would like an additional option of "Next X hours", ideally with customizable X

jgoldhammer:
I think that would be a nice feature... @leits WDYT?

KaiOelfke:
For me it would be even x minutes perhaps in 5 or 15 min intervals, but definitely below 1h.

igorkulman:
Which version is this feature available in? I am running the latest 3.2.1 released after the related PR has been merge but I do not see the option in the app.

leits:
Try this option in preferences:
<img width="812" alt="зображення" src="https://user-images.githubusercontent.com/12017826/113553546-9e018c00-9600-11eb-9ee2-1603f4256b13.png">

igorkulman:
Got it thanks, I was looking in the wrong place remembering the PR screenshot.

PR #186, opened 2021-02-12
Add setting to only show events starting in x minutes.

KaiOelfke:
### Status
**READY**

### Description
This PR adds a setting to only show events in x minutes. It implements the feature request #176. It's only filtering out future events that start later then the configured threshold. Current or past events are displayed as configured by other settings.

Bildschirmfoto 2021-02-12 um 19 08 41

1. Have the next event in e.g. 60 minutes.
2. Open Settings and configure it to show only events starting in 30 minutes.

The event should not be displayed in the title.

3. Move event to in 25 minutes.

The event should be now displayed in the title.


MeetingBar/Views/Preferences/AppearanceTab.swift
```diff
@@ -75,9 +75,7 @@ struct StatusBarSection: View {
             HStack {
                 Picker("Time", selection: $eventTimeFormat) {
                     Text("show").tag(EventTimeFormat.show)
-                    if #available(OSX 11.0, *) {} else {
```
jgoldhammer, line 78: issue : Please reintegrate the change from upstream. We want to hide this option in Big Sur...

MeetingBar/Views/Preferences/AppearanceTab.swift
```diff
@@ -140,14 +140,18 @@ struct EventsSection: View {
...
-
-
                 Picker("All day events:", selection: $allDayEvents) {
                     Text("show").tag(AlldayEventsAppereance.show)
                     Text("show only with meeting link").tag(AlldayEventsAppereance.show_with_meeting_link_only)
                     Text("hide").tag(AlldayEventsAppereance.hide)
                 }
             }
+            HStack {
+                Toggle("Show only events starting in", isOn: $showEventMaxTimeUntilEventEnabled)
```
jgoldhammer, line 150: **suggestion**: Can we change this to "Show only next event starting within next"? Or what do you think @leits ?

jgoldhammer: @leits question: Currently this option is in the events section. Should we move it to statusbar section to make clear that this only affects the statusbar and not the menu?

jgoldhammer (changes requested):
Overall looks good to me. Please lets discuss my questions .

MeetingBar/Views/Preferences/AppearanceTab.swift
KaiOelfke: I agree, it makes more sense in the status bar section and "Show only next event starting within" is clearer copy.

MeetingBar/Views/Preferences/AppearanceTab.swift
KaiOelfke: That change wasn't intended. Thanks for the heads up.

leits:
What worries me is that if there are no meetings at the next chosen time, the icon will be the same as when all the meetings are over. What do you think about it, @KaiOelfke, @jgoldhammer ?

jgoldhammer:
> What worries me is that if there are no meetings at the next chosen time, the icon will be the same as when all the meetings are over. What do you think about it, @KaiOelfke, @jgoldhammer ?

That would be indeed a problem for the user. It would mean to read all possible next event candidates from the apple calendar and show a different icon if there is another next event for today or today+tomorrow. Does this makes sense?

leits:
Yes, I think so.
Maybe the app icon?

KaiOelfke:
Maybe, I'll have time this weekend to look at this. There's also the SF Symbol: calendar.badge.clock

KaiOelfke:
I moved the logic to updateTitle and added an enum to handle the different cases. I'm not sure about the title logic. But the icon for the threshold case should be the app icon or `iconCalendar` and not the checkmark one.

MeetingBar/AppDelegate.swift
```diff
@@ -413,7 +427,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
             preferencesWindow.close()
         }
         preferencesWindow = NSWindow(
-            contentRect: NSRect(x: 0, y: 0, width: 700, height: 550),
+            contentRect: NSRect(x: 0, y: 0, width: 700, height: 600),
```
leits, line 430:
```suggestion
            contentRect: NSRect(x: 0, y: 0, width: 700, height: 550),
```

leits (approved):

PR #199, opened 2021-02-27
Feat/183 outlook safe urls

jgoldhammer:
### Status
**READY**

### Description
We will now recognize outlook safe links -> we will extract and decode the original link, so that the link can be handled with the original logic to detect a meeting link 

### Steps to Test

- Create a meeting with a safe url in it (see the original ticket)
- Check that the real meeting service is detected and shown up in the systembar and menu.


MeetingBar.xcodeproj/project.pbxproj
```diff
@@ -233,7 +230,6 @@
 		40DC7052250E5E1000217DD9 /* Frameworks */ = {
 			isa = PBXGroup;
 			children = (
-				14BAC84125C4AF1000C8A852 /* StoreKit.framework */,
```
leits, line 236: No need to remove this.)

MeetingBar.xcodeproj/project.pbxproj
```diff
@@ -589,6 +586,7 @@
 				COMBINE_HIDPI_IMAGES = YES;
 				CURRENT_PROJECT_VERSION = 114;
 				DEVELOPMENT_ASSET_PATHS = "";
+				DEVELOPMENT_TEAM = V8QGW6DP9U;
```
leits, line 589: Looks like development artifacts

MeetingBar/Helpers.swift
```diff
@@ -15,9 +15,42 @@ struct Bookmark: Encodable, Decodable, Hashable {
...
+ * The method will extract the real url from safe links and decode it, so that the following regex logic can detect the meeting service.
+ *
+ * The original link looks like this
+ *https://nam12.safelinks.protection.outlook.com/ap/t-59584e83/?url=https%3A%2F%2Fteams.microsoft.com%2Fl%2Fmeetup-join%2F19%253ameeting_[obfuscated]&data=[obfuscated]
+ *
+ * and the method will extract it to https://teams.microsoft.com/l/meetup-join/19%3ameeting_[obfuscated]
+ * If no m365 links are found, the original text is returned.
+ *
+ */
+fileprivate func cleanupOutlookSafeLinks( text: inout String) -> String {
```
leits, line 29: LGTM
But this is starting to be non-trivial and we need to start writing tests for link definition functionality.

jgoldhammer: You are right. I have much experiences with automated tests in other languages, but not with swift. How should we start?

jgoldhammer:
We should stash the commits from this PR when we merge it...

MeetingBar/Helpers.swift
leits: I will make a base for testing today, and then we will grow it.
But I'll merge this PR 👍

leits (approved):

PR #200, opened 2021-02-28
🎉feat (fantastical): allow to open calendar events in fantastical

jgoldhammer:
### Status
**IN DEVELOPMENT**

### Description
- we have added a menu item Open in fantastical to the event menu entry besides open in calendar
- the action will trigger the url handler for fantastical with a date and title
- fantastical will automatically open as app and show the appointment

- Create fantastical via homebrew (brew install --cask fantastical)
- Configure a calendar in Fantastical 
- Create an event which is available in meetingbar and fantastical
- open the event via "open in fantastical" -> fantastical should show the event

### Questions
- Should we introduce a new configuration option "Use calendar app" with 2 options "macos Calendar" and "Fantastical" so that the user can configure the used app?
- Should we check if fantastical is installed, so that fantastical and the whole configuration is only shown if the app is installed?


jgoldhammer:
@leits - what do you think about my questions?

leits:
Sorry for the delay, @jgoldhammer.

I think the best option, for now, will be:
1. not introduce the new configuration option
2. check if fantastical is installed and only then show "open in fantastical"

jgoldhammer:
Ok - totally fine for me. So user has two options: open in calendar and Open in Fantastical. Right?

leits:
> Ok - totally fine for me. So user has two options: open in calendar and Open in Fantastical. Right?
If fantastical is installed - yes

jgoldhammer:
> > Ok - totally fine for me. So user has two options: open in calendar and Open in Fantastical. Right?
> 
> If fantastical is installed - yes

I have integrated a check for the bundleid and tested locally with installed and removed fantastical...

leits (approved):

PR #202, opened 2021-03-01
🎉 feat(clipboard): allow to open meeting in clipboard

jgoldhammer:
### Status
**READY**

### Description
- we have introduced new functionality to access the clipboard
1. a new menu entry "Open meeting from clipboard"
2. open the link via shortcut

- define a shortcut for open link from clipboard
- Copy a link into the clipboard
- Use the shortcut or menu "open meeting from clipboard" 

image


jgoldhammer:
What do you think @leits ?

leits:
@jgoldhammer 
I think about this feature a lot.
The important thing in MB is privacy.
Therefore, I have a caveat against implicit monitoring, so I like the obvious action.
So LGTM.

It's out of scope, but I'm thinking of moving some actions to the "Quick Actions" submenu. 
It would be convenient to add a title hiding and skipping the next meeting there. WDYT?

MeetingBar/AppDelegate.swift
```diff
@@ -118,6 +118,31 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
...
+
+        if !clipboardContent.isEmpty {
+            let meetingLink = detectLink(&clipboardContent)
+            if meetingLink != nil {
+                openMeetingURL(meetingLink?.service, meetingLink!.url)
+            } else {
+                let validUrl = NSURL(string: clipboardContent)
+                if validUrl != nil {
+                    URL(string: clipboardContent)?.openInDefaultBrowser()
+                }
```
leits, line 138: Maybe send a message if we did nothing with the clipboard?

MeetingBar/AppDelegate.swift
```diff
@@ -118,6 +118,31 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
...
+    func openLinkFromClipboard() {
+        NSLog("Check macos clipboard for link")
+        var clipboardContent = getClipboardContent()
+        NSLog("Found \(clipboardContent) in clipboard")
+
+        if !clipboardContent.isEmpty {
+            let meetingLink = detectLink(&clipboardContent)
+            if meetingLink != nil {
+                openMeetingURL(meetingLink?.service, meetingLink!.url)
+            } else {
```
leits, line 134:
```suggestion
            if let meetingLink = meetingLink {
                openMeetingURL(meetingLink.service, meetingLink.url)
            } else {
```

leits:
@jgoldhammer, please, take a look at the comment.

jgoldhammer:
> @jgoldhammer
> I think about this feature a lot.
> The important thing in MB is privacy.
> Therefore, I have a caveat against implicit monitoring, so I like the obvious action.
> So LGTM.
> 
> It's out of scope, but I'm thinking of moving some actions to the "Quick Actions" submenu.
> It would be convenient to add a title hiding and skipping the next meeting there. WDYT?

Thanks for your opinion. Let us move this action when we work on the other feature to hide the title and skip next meeting.

MeetingBar/AppDelegate.swift
jgoldhammer: I have added a message.

MeetingBar/AppDelegate.swift
jgoldhammer: Changed.

leits (approved):

PR #215, opened 2021-03-18
Use UTF-16 for HTML tag stripping

### LINKED ISSUE #214: Bug: incorrect Japanese rendering in meeting notes
**App version**
3.1.2

**Describe the bug**
Html with Japanese renders incorrectly in the notes section of event details.
зображення

**To Reproduce**
1. Create an event with the description:
`<b><font>日本語テスト </font></b>`
2. Open event details in the app

0bmxa:
### Status
**READY**

### Description
Fixes #214 by using UTF-16 in the HTML tag stripping method.

I'm surprised Emojis in notes worked before, because I thought they (or some of them?) also need UTF-16, but maybe I'm wrong.

I don't know japanese, so I only tested it with the string in the issue: `<b><font>日本語テスト </font></b>`, and it looks good. Also regular HTML still works.


0bmxa:
Update: Tested it quickly with some longer text in Japanese and Korean (I just put the text from Issue 212 into Google Translate) and some more HTML, and the Japanese text is somewhat cut off at the end, but otherwise looks fine:
Screen Shot 2021-03-18 at 01 19 38
Screen Shot 2021-03-18 at 01 19 42

Note how the Japanese version is missing that line with `macOS [something] 11.2.2`. Not sure where that comes from. It's written into `item.attributedTitle` in `createEventItem(event:)` but it doesn't show up. Maybe some line terminator or so…?

leits (approved):

leits:
Thanks, @0bmxa!
I'll keep an eye on the case with `macOS [something] 11.2.2`.

PR #222, opened 2021-03-22
Mark notification as disabled only if authorization status is denied (Fix #219)

### LINKED ISSUE #219: Bug: The preferences has a “⚠️ Your macOS notification settings..” even though notifications for MeetingBar are enabled.
*from email*

Maybe it's because we only check for an Authorized state without taking Provisional or Ephemeral into account. 
https://github.com/leits/MeetingBar/blob/fd58e69906447d6724b3edeef7eb26c63a7814c7/MeetingBar/Views/Shared.swift#L27

leits:
### Status
**READY**

### DIFF
--- a/MeetingBar/Views/Shared.swift
+++ b/MeetingBar/Views/Shared.swift
@@ -24,7 +24,7 @@ struct JoinEventNotificationPicker: View {
 
         center.getNotificationSettings { notificationSettings in
             noAlertStyle = notificationSettings.alertStyle != UNAlertStyle.alert
-            notificationsDisabled = notificationSettings.authorizationStatus != UNAuthorizationStatus.authorized
+            notificationsDisabled = notificationSettings.authorizationStatus != UNAuthorizationStatus.denied
             group.leave()
         }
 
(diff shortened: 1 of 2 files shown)


leits:
Fix for #219

leits:
@jgoldhammer, do you have any concerns?

jgoldhammer (approved):

MeetingBar/Views/Shared.swift
```diff
@@ -24,7 +24,7 @@ struct JoinEventNotificationPicker: View {
 
         center.getNotificationSettings { notificationSettings in
             noAlertStyle = notificationSettings.alertStyle != UNAlertStyle.alert
-            notificationsDisabled = notificationSettings.authorizationStatus != UNAuthorizationStatus.authorized
+            notificationsDisabled = notificationSettings.authorizationStatus != UNAuthorizationStatus.denied
```
jgoldhammer, line 27: Is this correct? I think it must be == instead of !=

leits: You're absolutely right! Thanks!

leits: You're absolutely right!

Issue #221, opened 2021-03-21
Feature: Limit width of submenu items


mdbraber:
**Is your feature request related to a problem? Please describe.**
It seems that newlines are non-existent or ignored in notes field which therefore become extremely wide as submenu items. It would be great if there would be a max width to the submenu items.

**Describe the solution you'd like**
Max width for submenu items (either configurable or hardcoded)

**Describe alternatives you've considered**
Turning of the submenu items works, but does not allow access to relevant information

2Grey:
Check out PR [#266](https://github.com/leits/MeetingBar/pull/226)

jgoldhammer:
Thanks @2Grey !

PR #226, opened 2021-03-26
Add line splitting to newline

2Grey:
### Status
**READY**

### Description
Added line splitting to new line for long text in notes

Fix issue: #221


jgoldhammer:
Hi @2Grey ,

how can I test this? Creating an event with long notes text? Any specific I have to take care of?

2Grey:
@jgoldhammer, i tested it like this:
Created an event in the Calendar app in macOS with a very long notes text.
- For app without fix, it shows a sub-menu with width from screen edge to main menu view, also note text is cropped.
- For app with fix, it limit note text to 300px width and shows notes as multi-line.

I'm not sure how this will work for super long note text, when the screen height is not enough to display the text (but this is a rare case).

Also I have not tested this for the case where the text of the note contains a word longer than 300px width (maybe some German word).
I guess the menu will become this word wide.

2Grey:
@jgoldhammer, i checked some cases from previous comment:

- Long note in non-fixed app: [pic](https://imgur.com/NIE5Nrs)
- Same long note in fixed app: [pic](https://imgur.com/cCMvFo8)

- Super long note in fixed app: [pic](https://imgur.com/k6CHfBy)
It add a scroll to sub-menu.

- Super long word in note: [pic](https://imgur.com/t5MTSMj)
There is a problem, because word don't wrap;

jgoldhammer:
@2Grey - thanks for the examples. 
I have tested it a little bit and have increased the width to 450px. It looks fine in most cases 👍  
The following example has a teams link and this influences the UX for long notes.

image

What do you think @2Grey - can we have an easy solution for that?
What do you think @leits ? Should we merge it as it is and iterate later ? Or should we solve that first?

leits:
LGTM
It's definitely better than how it works now, let's merge it.

leits (approved):

PR #229, opened 2021-04-03
Feat/167 browser support

jgoldhammer:
### Status
**IN DEVELOPMENT**

### Description

This is the first version for browser management in Meetingbar.
It is not finished yet.

What works at the moment?
* add new browser configs
* edit a browser config
* delete a browser conifg
* add all browsers from the system
* delete all browser config
* choose a default browser for opening meeting links

What is missing?
* error handling for edge cases (e.g. delete browser config which is used
* UX polish

Out of scope at the moment:
* argument support for browser configs as arguments to pass over to browsers would require to run meetingbar not in  sandbox mode
* provide support to select a browser for each meeting service - we still have Google Meet, Teams and Zoom

### Some screenshots

image

image

* Open preference dialog and navigtate into browser tab
* Add new browsers and select the browser as generic


leits:
Wow. Great bunch of work!
What do you think about moving in modal under something like "configure browsers" button in Services tab?
I think a separate tab may overload Preferences window.

jgoldhammer:
I have integrated the configuration as modal view in services.

The services tab now looks like this - I have integrated the button "Configure browsers" which opens the modal dialog

image

The modal dialog now has an OK-button to close the modal dialog and return to the preferences dialog.

image

leits:
@jgoldhammer  Is it still *IN DEVELOPMENT* or we can merge it?

jgoldhammer:
Please test before we merge it. 

I currently see problems moving the browser position in the modal dialog - when moving one browser up or down, there are big spaces between the next browser config. Can you test this on your machine? Maybe we should deactivate it if this does not work. But this also shows up in the other dialogs, e.g. the bookmark dialog. Maybe a Big Sur bug?

On the other side, I still have not put the browsers as options for teams and zoom service because we have other option keys (not browser based) which will not work. What do you think what we should do here?

MeetingBar/AppDelegate.swift
```diff
@@ -412,29 +424,31 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
     @objc
     func createMeeting(_: Any? = nil) {
         NSLog("Create meeting in \(Defaults[.createMeetingService].rawValue)")
+        let browser: Browser = Defaults[.browserForCreateMeeting]
+
         switch Defaults[.createMeetingService] {
         case .meet:
-            openMeetingURL(MeetingServices.meet, CreateMeetingLinks.meet)
+            openMeetingURL(MeetingServices.meet, CreateMeetingLinks.meet, browser)
```
leits, line 431: Maybe make the browser not required and use Defaults[.browserForCreateMeeting] inside by default?

MeetingBar/Extensions/DefaultsKeys.swift
```diff
@@ -61,6 +61,11 @@ extension Defaults.Keys {
     // Bookmarks
     static let bookmarks = Key<[Bookmark]>("bookmarks", default: [])
 
+    // all browser configurations
+    static let browser = Key<[Browser]>("browser", default: [])
```
leits, line 65: Looks like "browsers", no?

MeetingBar/Views/Preferences/BrowserConfigView.swift
```diff
@@ -0,0 +1,254 @@
...
+                            Text(generatePath(browser: browser)).font(.system(size:11)).foregroundColor(Color.gray)
+                        }
+
+                        Spacer()
+
+                        Button(action: {
+                            self.browser = browser
+                            self.showingEditBrowserModal.toggle()
+                        }) {
+                            Image(nsImage: getPencilImage())
```
leits, line 43: What do you think about changing it to a built-in NSImage.touchBarComposeTemplateName instead of using custom image?

MeetingBar/Views/Preferences/BrowserConfigView.swift
```diff
@@ -0,0 +1,254 @@
...
+                .sheet(isPresented: $showingAddBrowserModal) {
+                    EditBrowserModal(browser: $browser)
+                }
+
+                Button(action: {
+                    self.showingDeleteAllAlert = true
+                }) {
+                    Image(nsImage: NSImage(named: NSImage.touchBarDeleteTemplateName)!)
+                    if #available(OSX 11.0, *) {
+                        //   image.offset(x: 0.0, y: 4.0)
```
leits, line 90: Commented?

MeetingBar/Views/Preferences/BrowserConfigView.swift
```diff
@@ -0,0 +1,254 @@
...
+                        alignment: .leading,
+                        spacing: 10
+                    ) {
+                        TextField("", text: $browser.name)
+                        TextField("", text: $browser.path)
+                        //    TextField("", text: $browser.arguments)
+                    }
+                    VStack(
+                        alignment: .leading
+                        //spacing: 10
```
leits, line 197: Comments?

MeetingBar/Views/Preferences/BrowserConfigView.swift
```diff
@@ -0,0 +1,254 @@
...
+
+                    Button(action: {
+                        self.presentationMode.wrappedValue.dismiss()
+
+                        let browserConfig = Browser(name: browser.name, path: browser.path, arguments: browser.arguments)
+                        browserConfigs.removeAll { $0.name == browser.name }
+                        browserConfigs.append(browserConfig)
+
+                        browserConfigs = browserConfigs.sorted { $0.path.fileName() < $1.path.fileName() }
+                        self.browser = Browser(name: "", path: "", arguments: "", deletable: true)
```
leits, line 221: Need to move this to separate function

MeetingBar/Views/Preferences/BrowserConfigView.swift
```diff
@@ -0,0 +1,254 @@
...
+                    }.disabled(self.browser.name.isEmpty || self.browser.name.isEmpty)
+                }
+            }.frame(width: 500, height: 200)
+            .padding()
+            .alert(isPresented: $showingAlert) {
+                Alert(title: Text("Can't add browser config"), message: Text(error_msg), dismissButton: .default(Text("OK")))
+            }
+        }
+
+        func chooseApplication() {
```
leits, line 233: Why "Application"?

leits:
Reproduced drag&drop bug. It worked fine for bookmarks recently, so I guess it's a Big Sur bug.

But I think we don't need a reordering here. Or there is some user need that I miss?

I think, we can extend Teams/Zoom options in separate PR.

MeetingBar/Extensions/DefaultsKeys.swift
```diff
@@ -74,7 +79,12 @@ extension Defaults.Keys {
     // custom url to create meetings
     static let createMeetingServiceUrl = Key<String>("createMeetingServiceUrl", default: "")
 
-    static let browserForMeetLinks = Key<Browser>("browserForMeetLinks", default: .defaultBrowser)
+    static let browserForMeetLinks = Key<Browser>("browserForMeetLinks", default: systemDefaultBrowser)
```
leits, line 82: Add backward compatibility for this change.

leits:
@jgoldhammer, today I noticed that Meet links open in the default browser, regardless of which option is selected.
I think we need to add tests for browser choosing.

leits:
Also there is no ` BrowserConfigView.swift` in `MeetingBar.xcodeproj/project.pbxproj`.

MeetingBar/Views/Preferences/AppearanceTab.swift
```diff
@@ -77,9 +77,7 @@ struct StatusBarSection: View {
             HStack {
                 Picker("Time", selection: $eventTimeFormat) {
                     Text("show").tag(EventTimeFormat.show)
-                    if #available(OSX 11.0, *) {} else {
-                        Text("show under title").tag(EventTimeFormat.show_under_title)
-                    }
+                    Text("show under title").tag(EventTimeFormat.show_under_title)
```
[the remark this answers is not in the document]
jgoldhammer: Fixd

MeetingBar/AppDelegate.swift
jgoldhammer: We are using openMeetingURL for other purposes as well, so it could be the wrong assumption to reuse a certain browser configuration in the method. WDYT?

MeetingBar/Extensions/DefaultsKeys.swift
jgoldhammer: yes- changed it.

MeetingBar/Views/Preferences/BrowserConfigView.swift
jgoldhammer: ok, is fine for me. Changed it.

MeetingBar/Views/Preferences/BrowserConfigView.swift
jgoldhammer: Removed it - was testing something. Good catch!

MeetingBar/Views/Preferences/BrowserConfigView.swift
jgoldhammer: Removed it. At first I wanted to support arguments for browser, but this does not work with the sandbox concept. So I removed it.

MeetingBar/Views/Preferences/BrowserConfigView.swift
jgoldhammer: Yes, I have now - fixed

MeetingBar/Views/Preferences/BrowserConfigView.swift
jgoldhammer: Changed to chooseBrowser- fixed.

jgoldhammer:
> Reproduced drag&drop bug. It worked fine for bookmarks recently, so I guess it's a Big Sur bug.
> 
> But I think we don't need a reordering here. Or there is some user need that I miss?
> 
> I think, we can extend Teams/Zoom options in separate PR.

I will remove ordering due to this bug.

jgoldhammer:
> Also there is no ` BrowserConfigView.swift` in `MeetingBar.xcodeproj/project.pbxproj`.

Added the file now. The problem is the project file which must be reset by me always because of the payment functionality where my local development certificate does not work with.

MeetingBar/Extensions/DefaultsKeys.swift
jgoldhammer: How should we do that here? This is a tricky one... Or do you have an idea?

jgoldhammer:
> @jgoldhammer, today I noticed that Meet links open in the default browser, regardless of which option is selected.
> I think we need to add tests for browser choosing.

Has been fixed with latest commit.

MeetingBar/AppDelegate.swift
leits: That makes sense. 👍

MeetingBar/Extensions/DefaultsKeys.swift
leits: We can check if the default value is a string on app startup. WDYT?

jgoldhammer: Can you please add this check? I have problems to get it right - maybe it is a good idea to use a different key browserForMeetLinksNew for the new option and keep the old browserForMeetLinks. WDYT?

leits (approved):

PR #241, opened 2021-04-15
✨feature: new visibility option for non all day events

jgoldhammer:
### Status
**READY**

### Description
We can now better control visibility of non all day events - following options are available:

- standard is show which alls all events, regardless the event contains a meeting or a link

image

- show as inactive without any link hides events in the statusbar when the event does not contain any link, but shows in the menubar as inactive - links in events now will also open in the browser

image

- show as inactive without any meeting hides events in the statusbar when the event does not contain any known meeting link, but shows in the menubar as inactive

image

- hide without any link hides events in the statusbar and menubar when the event does not contain any link

image

- hide without any meeting link hides events in the statusbar and menubar when the event does not contain any known meeting link

image

- Create 3 events (1 without any link, 1 with a known meeting service, 1 with a valid link, but no known meeting service)
- test the different options and check the result in menubar and statusbar


MeetingBar/Extensions/DefaultsKeys.swift
```diff
@@ -65,6 +65,9 @@ extension Defaults.Keys {
     // show all day events - by default true
     static let allDayEvents = Key<AlldayEventsAppereance>("allDayEvents", default: AlldayEventsAppereance.show)
 
+    // show all day events - by default show all, also events without any link
+    static let nonAllDayEvents = Key<NonAlldayEventsAppereance>("nonAllDayevents", default: NonAlldayEventsAppereance.show)
```
leits, line 69:
```suggestion
    static let nonAllDayEvents = Key<NonAlldayEventsAppereance>("nonAllDayEvents", default: NonAlldayEventsAppereance.show)
```

leits (dismissed):

jgoldhammer:
@leits - WDYT about this PR? Merge or not completely convinced?

leits (approved):

PR #242, opened 2021-04-15
🐛 fix(calendars): show read only calendars from delegateSources

jgoldhammer:
- after initialising the eventstore with the delegateSources a list of all calendars are displayed

### Status
**IN DEVELOPMENT**

### Description
The change shows the readonly calendars from the delegates in the macos calendar as well.

- create a delegate in macos calendar
- check the calendar tab in meetingbar that the calendar shows up


MeetingBar/Extensions/EventStore.swift
```diff
@@ -45,7 +45,7 @@ extension EKEventStore {
                 if Defaults[.allDayEvents] == AlldayEventsAppereance.show {
                     addEvent = true
                 } else if Defaults[.allDayEvents] == AlldayEventsAppereance.show_with_meeting_link_only {
-                    let result = getMeetingLink(calendarEvent)
+                    let result = getMeetingLink(calendarEvent, acceptAnyLink: false)
```
leits, line 48: Looks like changes from another PR

We definitely need more tests! 😄

leits:
LGTM
Thanks!

MeetingBar/Extensions/EventStore.swift
leits, line 48:
```suggestion
                    let result = getMeetingLink(calendarEvent)
```

leits (approved):

ckylape:
Is there a rough estimate when this might get wrapped into a release?

PR #243, opened 2021-04-15
Fix/181 incorrect time statusbar

jgoldhammer:
### Status
**READY**

### Description
Replaced NSBackgroundActivityScheduler with Timer implementation and placed the timer in the current runloop to also trigger the updates when the app is not in focus of the user

More details here:
https://www.hackingwithswift.com/articles/117/the-ultimate-guide-to-timer


leits:
Such a great and simple! Thanks!
I'll merge and run it locally for myself to test the changes.

leits (approved):

jgoldhammer:
@leits - thanks. Hopefully it will fix these problems.

2Grey:
@jgoldhammer, are you sure, that you should use 
```swift
let timer = Timer.scheduledTimer(timeInterval: 15, target: self, selector: #selector(updateStatusbar), userInfo: nil, repeats: true)
RunLoop.current.add(timer, forMode: .common)
```

Because when you init timer with `scheduledTimer` it add it to current RunLoop:
Description from Apple Doc:
> Creates a timer and schedules it on the current run loop in the default mode.

I think you should create it like this (without `scheduledTimer`):
```swift
let timer = Timer(timeInterval: 15.0, target: self, selector: #selector(updateStatusbar), userInfo: nil, repeats: true)
RunLoop.current.add(timer, forMode: .common)
```

jgoldhammer:
Great catch @2Grey ! I will create a PR for this ...

Issue #206, opened 2021-03-10
Feature: Temporarily hide meeting name


elainemckinley:
**Is your feature request related to a problem? Please describe.**
When in an impromptu screen sharing, sometimes I have a private/sensitive meeting coming up. It'd be nice to be able to temporarily hide the next meeting from within the menu bar instead of having to navigate to the preferences to do so.

**Describe the solution you'd like**
An option inside the menu bar (next to a meeting perhaps) to hide the meeting name. Some way of marking the meeting as sensitive so you see something like "Next meeting in 45m" instead when not clicking into the menu bar.

**Describe alternatives you've considered**
Currently I'm doing this through the preferences pane, which is fine but can be annoying.

**Additional context**
I've been looking for a project like this one for a while. I love it :)

jgoldhammer:
I love your feature request because I am often in the same situation. The menubar and statusmenu capabilities are very limited. Do you think we should add an entry in the menubar "Show/Hide meeting name" next to "Create meeting", "Join next meeting" and allow to bind it to a keyboard shortcut? What do you think @leits ?

leits:
Looks like a quick action in the menu.
Or option "hide the title during the meeting" in Preferences.
What do you think, @austinmckinley, @jgoldhammer?

austinmckinley:
My perfect wishlist for this would be a larger "current meeting" section at the top, with an item saying "Hide meeting name" or something that replaced the meeting name with "Current meeting ends in 10m" (if one is in progress) or "Next meeting in in 10m" (if not).
I'd imagine #205 could also make sense to put under this menu :)

jgoldhammer:
Hi @austinmckinley - I like your idea! The current approach with NSMenu and NSMenuItem is very limited, so I don not know if this looks good from a UX perspective.

What we could do is to show a seperate popup window when the statusitem is clicked with right mouse click via mouse or 2 fingers for touchpad. Currently this starts the next meeting, but we could instead show a seperate popup view with the current and upcoming meeting with the mentioned options. (hide meeting name, end meeting).

WDYT @austinmckinley  and @leits ?

elainemckinley:
I have no intentions of building this feature, but my gut feeling is that makes sense (although discoverability might be a little lower with that option). 

I also found another workaround (switching to Big Sur made me find a replacement to Bartender, so I can just drag MeetingBar to the hidden area now) so I'll leave the implementation up to y'all :)

leits:
It's done!

PR #254, opened 2021-04-22
Feat/206 Temporarily Hide Meeting Name

mbickers:
### Status
**READY**

### Description
Add functionality requested and discussed in #206 

- Menu item to show/hide meeting names
- Hides meeting names in notifications
- Adds keyboard shortcut selection for feature to preferences
- Russian localization for new strings

There is some overlap with the ability to change status bar title display in Preferences/Appearances, however, I think this feature is distinct because it is meant for temporary use (like wanting to hide an upcoming job interview while screen sharing) rather than permanent user preferences.


jgoldhammer:
Nice, looks fine. I have tested locally and everything works as expected (shortcut, statusbar, notifications)
I would merge the changes. WDYT @leits ?

jgoldhammer (approved):

PR #260, opened 2021-05-03
Fix casing in Zoom subdomains breaking regex

JamesMcMahon:
### Status
**READY**

### Description
My company has a cased subdomain whose meeting location were not being picked up by Meeting Bar. Follow up to https://github.com/leits/MeetingBar/discussions/234.

Use a prefix like `https://Test.zoom.us/j/` in the meeting location field. The prior regex would not pick this up.


leits (approved):

JamesMcMahon:
Is there a desire to make all the meeting urls case insensitive? If so, I can revise the approach to be more all encompassing.

_(If this is the wrong place to discuss I can also open a follow up issue.)_

leits:
Thanks, @JamesMcMahon.

So far, this is the first time the need has arisen.
Most services do not have a self-hosted version and those that do have usually use lowercase domains.

If there is at least one more case, then we can think about it.

PR #263, opened 2021-05-05
Display notes in NSTextView and make links clickable

jboulter11:
### Status
**READY**

Needs styling tweaks and probably some cleanup to your repo standards.

### Description
Adds clickable links to notes.  Uses an NSTextView to render the attributed strings.  Links are detected with a decent looking regex I pulled from stack overflow.

We can totally clean this up a bit more, and the text formatting isn't perfect yet, but I'd like your input there on what font and colors might match best.  The view itself seems to follow the same rendering as the attributed title of the menu item this replaced (see screenshot).  I added 10pt of padding to each side which seems to match up pretty well but can be tweaked.

Future work might be to pull links from HTML notes in the calendar events too (or just render the notes as HTML in the first place).

Run and navigate to a notes section with links in it!
Screen Shot 2021-05-05 at 1 54 48 PM

### DIFF
--- a/MeetingBar/Extensions/String.swift
+++ b/MeetingBar/Extensions/String.swift
@@ -150,3 +150,25 @@ extension String {
         return attributedString
     }
 }
+
+extension NSAttributedString {
+    func withLinksEnabled() -> NSAttributedString {
+        let linkDetectionRegexPattern = #"(http|ftp|https)://([\w_-]+(?:(?:\.[\w_-]+)+))([\w.,@?^=%&:/~+#-]*[\w@?^=%&/~+#-])?"#
+        guard let regex = try? NSRegularExpression(pattern: linkDetectionRegexPattern, options: .caseInsensitive) else {
+            return self
+        }
+
+        let newAttributedString = NSMutableAttributedString(attributedString: self)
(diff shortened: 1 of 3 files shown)


leits:
Wow. It's awesome, thank you!

Looks very good and I don't have any concerns about fonts and colors.

There are no comments on the formatting of the code either. I'm an amateur in swift as well as the vast majority of contributors. Therefore, the code may be a bit inconsistent, but at least it passes SwiftLint.

So would even be grateful if you could help us with good practices.  😄 

Just comment here when you're done and it can be merged.

leits (approved):

jboulter11:
Screen Shot 2021-05-06 at 12 30 34 PM

Just a few tweaks to layout (right padding), font size and color for better light/dark mode readability.  I also added tests on the link detection to at least cover some basic cases, more can be added easily as we find edge cases it doesn't cover.

I think we should also fix the font color on the other menu items but I'll do that separately when I get some more time.

jboulter11:
If you still have no issues with styling I think it's probably good to merge.

leits (approved):

leits:
Thanks, @jboulter11!

jboulter11:
Sorry, I forgot to run swiftlint, I'm used to it running itself. :P

leits (approved):

PR #318, opened 2021-08-20
Feat/317 no more meetings

jgoldhammer:
### Status
**READY**

### Description
It is now possible to set a label if no meeting will take place. The initial value is empty, so the old behaviour is kept by default. It is a opt-in...

The preference dialog:
image

The statusbar:
image

## Checklist
- [x] Localized
- [x] Added to changelog:
  - [x] [Changelog View](https://github.com/leits/MeetingBar/blob/master/MeetingBar/Views/Changelog/Changelog.swift)
  - [x] [CHANGELOG.md](https://github.com/leits/MeetingBar/blob/master/CHANGELOG.md)


jgoldhammer:
@leits  - swiftformat and swiftlint are conflicting each other. So when swiftformat is ok, swiftlint will complain about some broken rules. Can you please clarify? Is swiftlint more important?

jaklan:
@jgoldhammer I would rephrase the option name a bit, because in the current shape it's quite unclear what it actually does.

And it's also not obvious what should happen with the icon - if the custom label is set, then:
- we shouldn't use icon at all?
- we should use the existing "completed" icon?
- should that be dependent on the `Icon` option on the left side or not?

leits:
Sorry, @jgoldhammer, @jaklan, but I think it is unnecessary over-complication.
I don't see any other reason for this feature except personal preferences. And there is only one upvote on that ticket.

If I missed the point, feel free to reopen the pull and change my mind.

Issue #194, opened 2021-02-22
Open event url in browser if service not recognized


srstanic:
https://github.com/leits/MeetingBar/blob/a8065e7cdcfa51b201e8a1d7ef5c5a572300ceeb/MeetingBar/StatusBarItemControler.swift#L848

I'm curious why you decided to show a notification when the event URL is not recognized as one of the supported services instead of opening the event link in the browser? Opening the link in the browser is the default behavior of the `openMeetingURL()` function anyway. I find it limiting to not be able to click on the event and get the event URL opened in the browser. Would you consider changing this behavior or accepting a pull request?

jgoldhammer:
Thanks for the bug report. 
At the moment we are searching through notes, location and url field of the appointment for known meeting services. If we do not find any, we will shows this message. Changing this would mean to detect all kinds of links and guessing to open one of them. We are accepting pull requests!

Alternatively you could add a custom regex under Advanced to let meetingbar handle your meeting service. Maybe you have an example link for us?

srstanic:
I wouldn't call it a bug report, it's more of a UX improvement, if not even a preference. And I have to admit that my use case may be a bit out of the scope of the MeetingBar app, but here it is.

I schedule things to do in my calendar and sometimes these things have links. If they do, I put a link in the location field. Then when the time for this thing to do has come, I'd like to tap on the event in the MeetingBar to open the link. This is not strictly a meeting thing, more of a calendar thing. If I were designing this behavior, I would say that if a meeting service is not recognized, but there is at least one URL found in the event, pick the first one, and open it. Since you already check for links in the fields with the following order (location, url, notes), this would work perfectly for my use case and I believe it would be useful in case of unrecognized services as well. Certainly more useful than a notification about not recognizing a service.

Please let me know your thoughts! And thank you for a quick response!

jgoldhammer:
The use case for this app is to visualize and join meetings with one click. I guess this feature request goes beyond it, but would indeed bring a benefit. I will try to implement a solution for this...

jgoldhammer:
The link detection was part of another merge request which was merged. @leits removed this functionality for valid reasons (https://github.com/leits/MeetingBar/commit/efb157e5bb43ea0825e9d11ec9a0ac361f1bffe6#commitcomment-50292455). Maybe we can open this discussion again here to think about a useful integration. WDYT @srstanic about how this can be properly integrated?

srstanic:
[...]

PR #326, opened 2021-08-29
quick fix for issue #194

luckman212:
### Status
**READY**

### Description
Small change related to #194 that opens the `URL` of an event in the default browser if it's not handled by `getMeetingLink()`

I think this improves the end user experience without causing any harm or unexpected functionality.

1. Create an event that has just a simple URL e.g. `https://www.google.com/search?q=MeetingBar`
image
2. Click on this event from MeetingBar

### DIFF
--- a/MeetingBar/Helpers.swift
+++ b/MeetingBar/Helpers.swift
@@ -176,7 +176,6 @@ func detectLinks(text: String) -> [URL] {
 
 func openEvent(_ event: EKEvent) {
     let eventTitle = event.title ?? "status_bar_no_title".loco()
-    let eventUrl = event.url?.absoluteString ?? ""
     if let meeting = getMeetingLink(event) {
         if Defaults[.runJoinEventScript], Defaults[.joinEventScriptLocation] != nil {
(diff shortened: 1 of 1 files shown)


luckman212:
Is there anything I can (or need) to do @leits to request a review of this?

MeetingBar/Helpers.swift
```diff
@@ -189,6 +190,9 @@ func openEvent(_ event: EKEvent) {
             }
         }
         openMeetingURL(meeting.service, meeting.url, nil)
+    } else if eventUrl != "" {
+        // NSLog("opening \(eventUrl) in default browser")
+        event.url?.openInDefaultBrowser()
```
leits, line 195:
```suggestion
    } else if let eventUrl = event.url {
        eventUrl.openInDefaultBrowser()
```

Looks like you can do this all in one string.
Please check this!

leits (commented):
Hi @luckman212,
Thanks for the request!
Looks like makes sense, but please check my comment

MeetingBar/Helpers.swift
luckman212: Neat trick! @leits 
Looks like you're right of course. I committed that fix, tested & works.

leits (approved):

Issue #347, opened 2021-09-21
Question: Why does the current profile require In-App Purchase (StorerKit) support


fkrauthan:
I was wondering why the current build setup requires In-App Purchase entitlement to be enabled. It seems like just removing it works fine (as I am using a simple Developer account that does not support In-App Purchases entitlement).

PR #377, opened 2021-11-15
Removed in-app purchase entitlement as it does not seem to be required

fkrauthan:
### Status
**READY**

### Description
Removed the inapp purchase entitlement as it prevents local development unless you have a paid app store subscription. It also does not seem to be required as it compiles and runs perfectly fine without.

## Checklist
- [X] Localized - Not required
- [X] Added to changelog:
  - [X] [Changelog View](https://github.com/leits/MeetingBar/blob/master/MeetingBar/Views/Changelog/Changelog.swift)  - Not required
  - [X] [CHANGELOG.md](https://github.com/leits/MeetingBar/blob/master/CHANGELOG.md)  - Not required

Just compile the project. Only project file has changed


fkrauthan:
This would solve my question #347 as the answer seem to be there is no reason for this.

leits:
Good catch, @fkrauthan!

leits (approved):

PR #427, opened 2022-03-26
[DRAFT] Direct Google Calendar integration as data source

leits:
### Status
**IN DEVELOPMENT**

I started preparing a direct integration with Google Calendar at the beginning of winter. It was a private development request for implementing direct integration as an open-source part of MeetingBar.

At first, I thought it would be a separate release of a different code.
But I see more and more requests for such functionality, primarily for work-related needs. Many companies do not allow the integration of corporate calendars into the Apple ecosystem.

This version is fully operational and has passed a small closed beta test.

There are small pieces of functionality that need to be completed. Such as displaying events the next day.
But most of the work here is restoring integration with the macOS calendar, creating a UI to select and switch sources, and incorporating changes made to the main version of the app from December.

PR #434, opened 2022-04-03
[DRAFT] Direct Google Calendar integration as data source #427

leits:
### Status
**IN DEVELOPMENT**


MeetingBar/MeetingBar.entitlements
```diff
@@ -6,6 +6,10 @@
 	<true/>
 	<key>com.apple.security.files.user-selected.read-write</key>
 	<true/>
+	<key>com.apple.security.network.client</key>
+	<true/>
+	<key>com.apple.security.network.server</key>
```
jgoldhammer, line 11: Is server also necessary as entitlements? I thought client would be enough...

MeetingBar.xcodeproj/project.pbxproj
```diff
@@ -895,6 +939,30 @@
...
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/mxcl/PromiseKit";
+			requirement = {
+				kind = upToNextMajorVersion;
+				minimumVersion = 6.16.2;
+			};
+		};
+		14B08AD52749AFEE002E3892 /* XCRemoteSwiftPackageReference "SwiftyJSON" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/SwiftyJSON/SwiftyJSON";
```
jgoldhammer, line 960: The usage of SwiftyJSON could be a risk for the future. It is not maintained anymore and it seems that it has problems with compiling on M1. WDYT?

MeetingBar/AppDelegate.swift
```diff
@@ -226,277 +104,112 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
...
-            }
-        }
-
-        menuEventTitleLengthObserver = Defaults.observe(.menuEventTitleLength) { change in
-            if change.oldValue != change.newValue {
-                NSLog("Change menuEventTitleLengthObserver from \(change.oldValue) to \(change.newValue)")
-                self.statusBarItem.updateMenu()
-            }
+        // Settings change observers
+        appearanceSettingsObserver = Defaults.observe(keys: .statusbarEventTitleLength, .eventTimeFormat, .eventTitleIconFormat, .showEventMaxTimeUntilEventThreshold, .showEventMaxTimeUntilEventEnabled, .showEventDetails, .shortenEventTitle, .menuEventTitleLength, .showEventEndTime, .showMeetingServiceIcon, .timeFormat, .bookmarks, .eventTitleFormat) {
```
jgoldhammer, line 114: Thinking even more about it- at startup a lot of observers are getting fired. I do not know if we can improve performance to only update the title and menu once instead of multiple times. WDYT?

jgoldhammer: And we should think about a way to not fire any event even if the values has not been changed.

jgoldhammer (commented):
This looks very, very good! 
In general this opens the door for direct integration of other providers like M365 as well.

I will now test on my local machine and my google account.

jgoldhammer:
I do not get it to work locally unfortunately.

When fetching google calender, the api returns a 403 error code.

image


I have generated the clientID and clientSecret via Google API console and published the "app".

image

So clicking on changing the account triggers the oauth2 dialog from google:

image

MeetingBar/MeetingBar.entitlements
leits: Good catch! Borrow this part from some integration tutorial. 
I'll try to remove it and see if the solution still works.

MeetingBar.xcodeproj/project.pbxproj
leits: Another good catch. I'll try to use default swift methods for json parsing.

MeetingBar/MeetingBar.entitlements
leits: - [x] 03249db67115bdbd633e45b76a1a8c5b3373b7d7

MeetingBar.xcodeproj/project.pbxproj
leits: - [x] cf9d4e5ede33d39899891c5f20369d6029e422ff 

Looks a little dirtier, but quite working.

MeetingBar/AppDelegate.swift
leits: It turns out that the Defaults have a setting to simply turn it off. The interface is not very obvious, but very easy to implement.
docs: https://github.com/sindresorhus/Defaults#defaultsobserve

- [x] https://github.com/leits/MeetingBar/pull/434/commits/a84706d3b20015aaa724482bc1519b9c7638b249

leits:
@jgoldhammer, I made changes according to your comments, please take a look.

I'll merge these changes to integrate them with your other two pull requests.
But you can continue to review them here and I'll be grateful for your comments.

jimlesch:
this didn't work for me the first time I went through the authorization process. The second time I went through it I checked the last two boxes to authorize access to all calendars and now it is working as expected. Great job on this.

PR #470, opened 2022-06-15
Added support for optional values in meet links

quoimec:
### Status
**IN DEVELOPMENT**

### Description
Some google meet links appear to have some optional URL components that have no effect on the destination but that are missed by MeetingBar's regex pattern matching.

A normal URL: https://meet.google.com/mog-suqd-enn
A newly supported URL: https://meet.google.com/_meet/mog-suqd-enn

Both are valid and both remain supported but this PR allows meeting bar to recognise the second version.

## Checklist
- [x] Localized
- [ ] Added to changelog:

Add a calendar with the optional URL parameter listed above.

### DIFF
--- a/MeetingBar/MeetingServices.swift
+++ b/MeetingBar/MeetingServices.swift
@@ -233,7 +233,7 @@ func openMeetingURL(_ service: MeetingServices?, _ url: URL, _ browser: Browser?
 }
 
 struct LinksRegex {
-    let meet = try! NSRegularExpression(pattern: #"https?://meet.google.com/[a-z-]+"#)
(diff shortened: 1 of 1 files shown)


quoimec:
Hello, code-wise, this PR is ready to go, I just haven't added it to the changelog yet as I am a bit confused as to what I should do here as there are multiple versions. The project is at version 4.0.5, the changelog view is at 4.0.0 and the CHANGELOG.md is at 3.11.0. Happy to clean these up but would appreciate some direction for what I should do here. Should I create a new version in all three at 4.0.6?

leits (approved):

PR #471, opened 2022-06-16
Added new preference for controlling bookmark collapse

quoimec:
### Status
**IN DEVELOPMENT**

### Description
Added a single setting for controlling the threshold under which the Bookmarks menu collapses to a submenu. This was previously always 3 but can now be configured by the user (default remains 3). This has been a personal gripe of mine for a while as I would rather just have quick access to all of my bookmarks rather than needing a submenu. 

## Checklist
- [ ] Localized
- [ ] Added to changelog:

### DIFF
--- a/MeetingBar/AppDelegate.swift
+++ b/MeetingBar/AppDelegate.swift
@@ -315,7 +315,7 @@ class AppDelegate: NSObject, NSApplicationDelegate, UNUserNotificationCenterDele
             preferencesWindow.close()
         }
         preferencesWindow = NSWindow(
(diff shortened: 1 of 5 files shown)


quoimec:
Hello, I have also just opened https://github.com/leits/MeetingBar/pull/470 but had a question on that about what I should do re: versioning. I have another question for this PR though as I have not contributed to a localised project before, what is the process for doing the translation of text? Is using Google Translate sufficient?

leits:
Thanks for the pull request, @quoimec!

I've been thinking about this feature for a while. Since there is no support from other users of the app, I prefer not to add more customization to the already huge settings options.
Sorry!

PR #501, opened 2022-09-07
Add preference to show events for tomorrow in the status bar

0ui:
### Status
READY

### Description
I was facing a problem due to the fact that I work with an organization on the other side of the world and my meetings usually start at midnight. This makes those events part of "tomorrow" which will never show up in the status bar so MeetingBar isn't useful prior to my first meeting. I can show events for "today and tomorrow" and set a range but it will not display a meeting that happens in 30 minutes if that meetings occurs past midnight.

#### Solution
My solution is to add a new preference that conditionally appears when events are shown for tomorrow. In this demo you can see:

1. I cannot see my meeting within the next hour
2. I show events for "today and tomorrow" and the meeting appears within the menu but not in status bar
3. I activate the new preference and now I can see my meeting coming up in the status bar

https://user-images.githubusercontent.com/734006/188869273-d031b778-0a0e-45a1-bb95-3a2ccb033173.mov

I realize this feels a bit awkward being outside the "status bar" area of preferences but

1. the "events" section does filter which events show in the status bar
2. this preference is only valid when events are shown for both today and tomorrow
3. It should not affecting existing behavior

I am not a Swift/MacOS developer so I apologize in advance for any rookie mistakes. I'm happy to update this with a different solution if this one isn't acceptable.

## Checklist
- [ ] Localized
- [ ] Added to changelog:

1. Create an event at midnight/00:00/12AM
2. Set system time to 30 minutes prior
3. Observe that the event is not visible in the status bar

### DIFF
--- a/MeetingBar/Extensions/DefaultsKeys.swift
+++ b/MeetingBar/Extensions/DefaultsKeys.swift
@@ -22,6 +22,7 @@ extension Defaults.Keys {
     static let onboardingCompleted = Key<Bool>("onboardingCompleted", default: false)
 
     static let showEventsForPeriod = Key<ShowEventsForPeriod>("showEventsForPeriod", default: .today)
+    static let includeTomorrowInStatusBar = Key<Bool>("includeTomorrowInStatusBar", default: false)
     static let joinEventNotification = Key<Bool>("joinEventNotification", default: true)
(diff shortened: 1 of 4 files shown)


marcusziade (approved):
LGTM, nice work 👍

leits:
Hi @0ui,
Thanks for your work!
I've reverted the previous behavior by showing the next meeting title from tomorrow for a two-day view. So now there is no need to add an extra setting.

PR #624, opened 2023-06-07
Fix event start script bug

benbernard:
### Status
**READY**

### Description
Found a bug with the event start script setting.  It will only run if a join script is also setup, I think due to just a type in the accessing Defaults

## Checklist
- [x] Localized (none required)
- [ ] Added to changelog: (unclear to me if you want this given that I didn't want to make a whole version, happy to do what you want here)

The bug happens if you have the run event script set but no join event. So Advanced settings looks like this:
image

In order for that to work, you have to have the join event location also set, which means you have to have that turned on.  This fix changes to check the event start script location, which is what I believe was intended.


benbernard:
I did test it! Thanks for merging

PR #632, opened 2023-07-07
Change Localizable.strings "あと" to "まで".

shu-pf:
Change to correct Japanese grammar.

### Status
**READY**

### Description

Thank you for the great software.
I am Japanese and I use it everyday.

I have one problem in using it.
The time until the meeting starts is expressed as "あと", but "あと" gives the nuance that the meeting is in progress.

Therefore, I was worried several times that the meeting had already started.

So I submit this PR.

Thank you very much for making the software.

## Checklist
- [x] Localized


leits (approved):

leits:
Thank you, @shu-pf!

Issue #888, opened 2026-02-01
Bug: Crashes on Outlook (MacOS Calendar) Integration


ryedog:
### macOS version

15.7.3

### MeetingBar version

4.11.6

### Installation source

AppStore

### Calendars provider

macOS Calendar app

### Bug description

After several days of usage, the MeetingBar menu bar icon disappears even though the MeetingBar process continues running. The app becomes unresponsive and cannot be recovered without removing the Outlook account from macOS Calendar.

I have found posts indicating MeetingBar may encounter issues if there is a corrupted, malformed, or otherwise unexpected calendar event, causing the app to hang.

I’m happy to help troubleshoot (logs, sample events, enabling debug mode, etc.) if it helps identify the root cause. Please let me know how I can help.

Issue #796, opened 2025-01-31
Bug: MeetingBar hangs at launch or shortly


537m3:
### macOS version

15.3 (24D60)

### MeetingBar version

4.10.0

### Installation source

AppStore

### Calendars provider

Google Calendar

### Bug description

Has been working flawlessly for over a year. About a week ago it started hanging, usually right on launch, but occasionally after running for a short period of time (has run for over 30 mins once since the symptoms began, but lately hangs immediately at launch before even displaying any upcoming meetings in the right menubar location. The icon itself does display, and I get the spinning pinwheel when I hover over it. Activity Monitor shows it as Not Responding, generally with less than 1s of cumulative CPU time. Continues until I force quit.

I tried rebooting several times and reinstalling the app. Same symptoms.

Unsure if related, but Apple's Single Sign-On app (which is launched automatically, not by me) is also always Not Responding in Activity Monitor and similar very low CPU time before hanging and never recovering. 

Is there any way I can get log output for the app to see what it's doing before it hangs?

### Reproduction steps

1. Launch MeetingBar
2. Notice it is hung and unusable

### Expected behavior

Normal operation

virtue3:
Also seeing the same behavior.  

Can't really do anything since the app immediately crashes off the menu bar.

I tried switching from google calendar to mac calendar and same behavior.

danieldisu:
Happening to me too

I started seeing it on 15.3.1

wolvgrad98:
Also happening in MacOS 15.3.2.

bzenker-amplify:
> Can't really do anything since the app immediately crashes off the menu bar.

I wonder if this deserves it's own issue as the OP doesn't have this behavior, but personally this is the behavior on my side. The app is **not** hanging, but it does **not** display on the Apple menu bar.

cesarferradas:
Can confirm this started happening since MacOS 15.1 for me too.

leits:
Hi folks,
Thanks again for reporting this performance issue. I’ve just published a 4.11 beta release with a number of under-the-hood performance optimizations in MeetingBar. You can grab it here:
https://github.com/leits/MeetingBar/releases/tag/v4.11.beta 

Once you’ve had a chance to try it, could you please let me know whether you still see the issues? Any logs, profiled timings, or specific repro steps you can share will be super helpful. Thanks!

glenjamin:
I've just tried switching to the beta, but I'm still getting a crash on startup. This is the crash report from `Console`

[...]

Issue #791, opened 2025-01-14
Bug: Google Calendar truncates links, causing Zoom URLs with prefilled passwords to not work


andrewmin:
### macOS version

15.2 (24C101)

### MeetingBar version

4.10.0

### Installation source

AppStore

### Calendars provider

macOS Calendar app

### Bug description

For certain instances, when I click to join a Zoom meeting, it pops up "Enter meeting passcode" even though the URL in the calendar invite has a prefilled passcode.

I believe what is happening is the following. When a Zoom user sends an invite through Google calendar, it includes two URLs, which seem to be a rich text and a plain text:

![Screenshot 2025-01-14 at 2 05 51 PM](https://github.com/user-attachments/assets/e05724b4-3591-4c7d-8f6e-ef686730d9da)

The rich-text URL appears to truncate (both the HTML and displayed text). Here is what shows up in MacOS Calendar (which does not render HTML) for the rich text link (the plain text link renders fine, as seen above):

<img width="246" alt="Screenshot 2025-01-14 at 2 08 28 PM" src="https://github.com/user-attachments/assets/30c00a67-cc6d-4c0a-8d09-c1bcc1f94cfc" />

I'm not sure why this is but wonder if we can use some sort of logic to look for all relevant Zoom links and if one has a password and one does not, use that password.

As a note (probably obvious from attached) we use URLDefense / Proofpoint which may be why it wraps.

Issue #776, opened 2024-11-12
Bug: Heavy power usage and hangs


timcappalli:
### macOS version

14.7 (23H124)

### MeetingBar version

4.10.0

### Installation source

Homebrew

### Calendars provider

Google Calendar

### Bug description

MeetingBar is drawing significant amounts of power over the past month or so.

I've also noticed that when I hover or click on the menu bar item, it often hangs / pinwheels for a good 3-5 seconds. Not sure if it is related.

### Screenshots/screen recordings

<img width="738" alt="image" src="https://github.com/user-attachments/assets/ffbba1f3-c9d1-4626-b205-270ea01f17fe">

leits:
Hi folks,
Thanks again for reporting this performance issue. I’ve just published a 4.11 beta release with a number of under-the-hood performance optimizations in MeetingBar. You can grab it here:
https://github.com/leits/MeetingBar/releases/tag/v4.11.beta 

Once you’ve had a chance to try it, could you please let me know whether you still see the issues? Any logs, profiled timings, or specific repro steps you can share will be super helpful. Thanks!

leits:
Thanks for reporting. The bug fix has been released!
Feel free to reopen the issue if the error still occurs.

PR #936, opened 2026-06-24
Fix infinite loop in Outlook SafeLink cleanup

jamtur01:
## What's going on

I ran into MeetingBar getting stuck on "Initializing" in the Calendars view on 5.0.0 — no events, and Refresh did nothing. After a fair bit of digging it turned out to be a single calendar event sending `cleanupOutlookSafeLinks` into an infinite loop, which quietly wedges the whole calendar sync.

## Why it happens

`cleanupOutlookSafeLinks` re-matches the first SafeLink each pass and rewrites it to its decoded `url=` target, repeating `while !links.isEmpty`. The loop doesn't guarantee it makes progress:

- If the captured `url=` value isn't valid percent-encoding, `removingPercentEncoding` returns `nil`, the text is never rewritten, and the same match is found again on the next pass — round and round forever.
- Same thing if a rewrite doesn't actually change the text.

The capture stops at the first whitespace, so a meeting URL that's wrapped or truncated in the middle of a `%XX` escape leaves a dangling/incomplete escape, which is exactly what trips `removingPercentEncoding`.

Since the event build runs on a background task, the loop doesn't beachball the UI — it just spins a core and the menu sits on "Initializing". (On older versions, when this ran on the main thread, it showed up as a beachball / high CPU instead.)

This isn't new — it goes back to #199 (2022), so any version with an affected event could hit it.

## The fix

Make the unwrap loop always move forward: stop if the target can't be decoded or if a rewrite leaves the text unchanged, with a hard cap as a backstop. Valid and nested SafeLinks still unwrap exactly as before.

## Tests

Added regression coverage in `MeetingLinkDetectorTests`: an undecodable target, a self-referential SafeLink, a normal SafeLink, nested SafeLinks, and an event that carries a malformed SafeLink next to a valid link. All `MeetingBarLogic` tests pass via `swift test`.

## Possibly related

This might be behind a few existing reports, though I can't say for sure:
- #888 — "becomes unresponsive" with an Outlook account; the reporter even guessed a malformed event was causing a hang. Would be worth checking with them.
- #478 (one appointment pegging 100% CPU) and the #762 / #776 / #796 / #671 hang/freeze cluster look like they could be the same thing.

## Not included

Events whose SafeLink target is genuinely truncated still won't surface a join link — the URL text past the wrap is just gone (similar theme to #791). Recovering those would be a separate change; this PR is only about not hanging.

## A note on the branch

I branched off `master`. Not sure if `v5` is the right base here — happy to do either, just let me know.

## Checklist
- [x] Tests added
- [ ] Localised — no user-facing strings
- [ ] Changelog — happy to add an entry if you'd like

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Outlook SafeLinks to safely and deterministically unwrap redirect targets, including malformed/undecodable payloads.
  * Added safeguards to prevent non-terminating rewriting loops while ensuring valid meeting links in the same text are still detected correctly.
* **Tests**
  * Added regression tests covering SafeLink unwrapping, malformed `url=` handling, self-referential cases, and end-to-end meeting link detection with mixed valid and invalid links.

### DIFF
--- a/MeetingBar/Meetings/MeetingLinkDetector.swift
+++ b/MeetingBar/Meetings/MeetingLinkDetector.swift
@@ -220,6 +220,23 @@ func detectMeetingLink(_ rawText: String, customRegexes: [String] = []) -> Meeti
     return nil
 }
 
+/// Rewrites Outlook SafeLink wrappers in `rawText` back to their real
+/// targets, so meeting-link detection sees the underlying URL rather than the
+/// `…safelinks.protection.outlook.com/…url=<encoded>` redirect. Each pass
+/// unwraps the first SafeLink and re-scans, which also resolves SafeLinks
+/// nested inside one another.
+///
+/// The loop is deliberately bounded and only continues while it makes forward
+/// progress. It stops when:
+///
+/// 1. the `url=` value can't be percent-decoded — a malformed or truncated
+///    `%` escape, which `removingPercentEncoding` reports as `nil`;
+/// 2. a rewrite leaves the text unchanged, so the same match would be found
+///    again; or
+/// 3. a hard pass cap is reached, as a final backstop.
+///
+/// Without these guards a single event carrying a malformed SafeLink would
+/// loop forever and wedge calendar sync.
 func cleanupOutlookSafeLinks(rawText: String) -> String {
     guard let outlookSafeLinkRegex else { return rawText }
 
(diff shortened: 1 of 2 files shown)


leits (approved):

leits:
Thanks @jamtur01!

PR #937, opened 2026-06-24
Fix empty edit box when editing a custom regex

jamtur01:
Clicking **edit** on a custom regex (either the "filter out meetings" or "meeting link" list) opened an empty box instead of the pattern I clicked on — so I had to retype the whole thing to make a small change.

The editor was presented with `.sheet(isPresented:)`, whose content gets built once while the draft is still empty and then reused, so the real value never made it in. Switching both sections to `.sheet(item:)` rebuilds the editor with the draft I actually picked, and `RegexEditDraft` is now `Identifiable` to drive that.

After the change: **Add** opens empty as before, **edit** pre-fills with the existing pattern, and the two no longer step on each other.

No logic changed — this is purely the preferences view wiring, so the existing `RegexListEditingPolicy` tests still pass. There is no UI-test target to cover the presentation itself, so I verified it by hand in a local build.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the add/edit flow for regex entries in Preferences so the correct edit dialog opens consistently.
  * Updated how the edit draft is managed, reducing mix-ups when switching between adding and editing.
  * Ensured the edit dialog seeds and refreshes its text reliably when presented.
  * Fixed save behavior so it only proceeds when a valid draft is available.

### DIFF
--- a/MeetingBar/Preferences/AdvancedTab.swift
+++ b/MeetingBar/Preferences/AdvancedTab.swift
@@ -251,8 +251,7 @@ struct NSScrollableTextViewWrapper: NSViewRepresentable {
 struct FilterEventRegexesSection: View {
     @Default(.filterEventRegexes) var filterEventRegexes
 
-    @State private var showingEditRegexModal = false
-    @State private var regexDraft = RegexEditDraft.adding()
+    @State private var regexDraft: RegexEditDraft?
 
     var body: some View {
         DisclosureGroup("preferences_advanced_event_regex_title".loco()) {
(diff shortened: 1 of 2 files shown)


jamtur01:
> 🧹 Nitpick comments (1)
> > MeetingBar/Preferences/AdvancedTab.swift (1)> `407-435`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _💤 Low value_
> > > **Optional: seed `draftRegex` in `init` to avoid a one-frame empty field on edit.**
> > > With `.sheet(item:)` the modal gets a fresh view identity on every presentation, so `@State(initialValue:)` in `init` now seeds correctly before the first render. As written, `draftRegex` starts empty and is only set in `.onAppear`, so an edit briefly shows an empty `TextField` (and a momentarily disabled Save button) before the seed value appears.
> > > 

Addressed this and re-pushed.

leits (approved):
Thanks @jamtur01!

Issue #369, opened 2021-11-09
Feature: Support for Zhumu/WeMeeting


hbarthels:
It would be great if you could add support for Zhumu/WeMeeting.

The links look like this: `https://welink.zhumu.com/j/XXXXXXXXX?pwd=YYYYYYYYYYYYYYYYYYYYYYYYYYYYYYYY`, where `X` are numbers and `Y` are letters of numbers. I don't know if the length of those sequences is fixed.

The corresponding app to participate in those meetings is called WeMeeting.

PR #939, opened 2026-06-28
Fix Zhumu meeting link detection

jamtur01:
I was playing around with all the meeting-link recognitions and noticed this one's regex looked a bit odd: `[0-9]+?pwd=`. That `?` isn't a literal query separator — it's a lazy quantifier on the digits — so the pattern only matches `/j/123pwd=abc`, which isn't a URL Zhumu actually produces. Real links carry a `?` (`/j/123?pwd=abc`) and slip straight through. It's been this way since Zhumu support first landed in 3d9e841 (resolving #369).

I did a bit of digging into what Zhumu links look like in the wild and broadened the pattern to the real formats:
- `/j/<id>` with an optional `?pwd=` passcode
- the web-client `/wc/join/<id>?wpk=` links

Added detector tests for all three, plus a couple confirming non-meeting Zhumu pages still don't match.

**Big caveat:** I've honestly never used Zhumu myself *shrug*, and I'm going purely off links I found online — so I might be misunderstanding the syntax. And given it apparently never worked and nobody reported it, I'm genuinely not sure this is worth fixing. If I've got the format wrong or you'd rather leave it, no hard feelings — feel free to close. :)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved meeting link detection for Zhumu/WeMeeting by recognizing more join URL variants and handling optional query parameters more accurately.
  * Reduced false positives by rejecting non-meeting Zhumu pages and incomplete/invalid join link formats.
* **Tests**
  * Added unit tests covering supported `/j/...` and `/wc/join/...` links (including required/optional query parameters) and ensuring invalid URLs return no result.

### DIFF
--- a/MeetingBar/Meetings/MeetingProvider.swift
+++ b/MeetingBar/Meetings/MeetingProvider.swift
@@ -454,7 +454,7 @@ extension MeetingProvider {
             make(
                 .zhumu,
                 icon: "zhumu_icon",
-                pattern: #"https://welink\.zhumu\.com/(?:j|wc/join)/[0-9]+(?:\?(?:pwd|wpk)=[a-zA-Z0-9]+)?"#),
+                pattern: #"https://welink\.zhumu\.com/(?:j/[0-9]+(?:\?pwd=[A-Za-z0-9]+)?|wc/join/[0-9]+\?wpk=[A-Za-z0-9]+)"#),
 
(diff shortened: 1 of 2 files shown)


leits (approved):

leits:
Thanks @jamtur01!

PR #945, opened 2026-07-04
Don't let a YouTube link in the notes beat the real meeting link

jamtur01:
### Status
**READY**

### Description
Fixes a link-ranking regression from the candidate-based detection work: `MeetingLinkCandidatePolicy` breaks same-source-priority ties by raw URL length across *all* services, so a longer incidental link in an event's notes (e.g. YouTube) beats the actual meeting link (e.g. a shorter Google Meet URL) for the menu bar icon, the join action, and auto-join.

The length tiebreak came in with e6460c0 for the Zoom `?pwd=` truncation case, back when each source field produced at most one candidate — different services could never tie. It became reachable across services in 9199682, when collection started gathering every regex match per source.

`best` and `ranked` now share a single comparator: source priority first (unchanged), then `MeetingServices.allCases` catalogue order when services differ (a `nil` service ranks last), and URL length only between candidates of the same service — preserving the original Zoom-password intent.

## Checklist
- [ ] Localized — n/a, no user-facing strings changed
- [ ] Added to changelog — n/a, fixes an unreleased regression (introduced during 5.0 development)

1. Create a calendar event whose notes contain both a YouTube link and a shorter Meet link, e.g. `https://www.youtube.com/watch?v=dQw4w9WgXcQ&list=PL2026-team-agenda-recordings` and `https://meet.google.com/abc-defg-hij`.
2. Without this change, the status bar and join action target the YouTube URL; with it, the Meet link wins. The "open with another link" list also orders Meet ahead of YouTube.
3. `swift test --filter MeetingLinkCandidateTests` — 18 tests: 3 new regression tests (watched fail against the old comparator), a pin that provider conference data still beats notes across services (mutation-tested), and the pre-existing Zoom-truncation test guarding the same-service length rule.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved meeting-link selection with consistent ranking across source priority, conferencing service, match position, and URL details.
  * More reliably distinguishes known conferencing services from generic links and preserves match context when rewriting Meet URLs.
  * Recognizes and decodes nested Google Calendar redirects and Outlook SafeLinks before selecting meeting links.

* **Tests**
  * Expanded coverage for candidate selection, deduplication, tie-breaking, redirects, and alternate-join ordering.

* **Documentation**
  * Updated the meeting-link ranking description to reflect current behavior.

### DIFF
--- a/MeetingBar/Meetings/MeetingLinkDetector.swift
+++ b/MeetingBar/Meetings/MeetingLinkDetector.swift
@@ -151,30 +151,44 @@ enum MeetingLinkCandidatePolicy {
     /// Picks the best candidate for an event:
     ///
     /// 1. by source priority — provider conference data beats notes;
-    /// 2. within the same source, the longer URL wins so a Zoom link that
+    /// 2. within the same source, catalogue order (`MeetingServices.allCases`)
+    ///    decides between different services, so an incidental YouTube link in
+    ///    the notes cannot shadow the real Meet link;
+    /// 3. within the same service, the longer URL wins so a Zoom link that
     ///    carries a password/token suffix beats a truncated form of the
     ///    same URL found in another source slot.
     static func best(from candidates: [MeetingLinkCandidate]) -> MeetingLinkCandidate? {
-        candidates.max { lhs, rhs in
-            if lhs.source.priority != rhs.source.priority {
-                return lhs.source.priority < rhs.source.priority
-            }
-            return lhs.url.absoluteString.count < rhs.url.absoluteString.count
-        }
+        candidates.min(by: isRankedBefore)
     }
 
     /// Returns candidates ranked best-to-worst, deduplicated by URL string.
     /// Useful for a "open with another link" menu without re-running detection.
     static func ranked(from candidates: [MeetingLinkCandidate]) -> [MeetingLinkCandidate] {
-        let unique = Dictionary(grouping: candidates, by: { $0.url.absoluteString })
+        Dictionary(grouping: candidates, by: { $0.url.absoluteString })
             .compactMapValues { best(from: $0) }
             .values
-        return Array(unique).sorted { lhs, rhs in
-            if lhs.source.priority != rhs.source.priority {
-                return lhs.source.priority > rhs.source.priority
-            }
-            return lhs.url.absoluteString.count > rhs.url.absoluteString.count
+            .sorted(by: isRankedBefore)
+    }
+
+    /// True when `lhs` outranks `rhs`. Candidates with a `nil` service rank
+    /// after every catalogued service.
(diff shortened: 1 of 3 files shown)


leits:
@jamtur01! Thanks for digging into this! The regression is real, and keeping the source priority and Zoom `?pwd=` behavior makes sense.

I don’t think we should use `MeetingServices.allCases` for ranking, though. Its order isn’t a real priority list. It fixes Meet + YouTube because Meet happens to come first, but YouTube would still beat Slack, LiveKit, Cal.com, Riverside, and any new service added later.

I’d make the ranking work like this:

1. Higher-priority source wins, as it does now.
2. Within the same source, conferencing links beat content links such as YouTube.
3. For two links from the same service, the longer URL wins, preserving the Zoom `?pwd=` case.
4. For different services in the same tier, keep the order in which the links appear in the event field.

That last part probably means keeping the match position on the candidate instead of relying on the order of `MeetingServices.allCases`.

Please also add a test with YouTube and a service currently declared after it, Slack would be a good example, plus one test through `MeetingLinkDetector.detect`.

There’s also a comment lower in `MeetingLinkDetector` that still describes URL length as the only same-source rule.

Thanks, this is very close!

jamtur01:
Yeah, that's a fair point; `allCases` is a bit magical. Here's what I'm thinking, basically your four steps:

1. Source priority stays exactly as it is.
2. Within a source, split services into tiers; real conferencing links beat content links (YouTube, Vimeo). I'd do this with a small explicit tag on the service rather than any ordering, so it's obvious what's what and adding a service doesn't accidently change ranking.
3. Same service → longer URL still wins, keeps the Zoom ?pwd= case.
4. Same tier, different services → earliest match in the field wins. That means carrying the match position on the candidate like you said, since right now the collection order comes from `allCases` and not where the link actually appears in the text.

One extra thing I noticed while poking at it: `.url ("Any Link")` and `.other` have no regex, so they never really compete from scanning, but the policy still needs untagged/nil candidates to rank last, which `allCases` was quietly doing for us via the `Int.max` fallback. So I'm leaning toward three tiers rather than two: conferencing → content → generic/unknown. Keeps "a real service beats a catch-all" working instead of leaving it to a string tiebreak.

Tests: I'll add YouTube + Slack (Slack's declared after YouTube, so it fails the old comparator and passes the new one), a same-tier ordering one, keep the Zoom truncation test, and add a full run through `MeetingLinkDetector.detect`. And I'll fix that stale comment that still says length is the only same-source rule.

Does the three-tier split sound right to you, or would you rather keep it to just conferencing vs content and handle the catch-all some other way?

leits:
The three-tier approach sounds good to me. Good catch!

PR #946, opened 2026-07-04
Give EventKit occurrences their own ids so dedup stops eating them

jamtur01:
### Status
**READY**

### Description
In the today+tomorrow view, a daily recurring meeting only shows today's instance — tomorrow's occurrence silently disappears from the menu, next-event selection, and notifications.

Cause: the id-keyed dedup added in 34293d6 (via #931) collapses events by `MBEvent.id`, and the EventKit adapter maps that id from `calendarItemIdentifier` — which EventKit reuses for every occurrence of a recurring event. So occurrences of the same series look like duplicates and get collapsed into one.

Before touching it I went back to 34293d6 to check what it was protecting against: the same event landing in a fetch more than once and double-notifying. Those true duplicates share both the item identifier *and* the start date.

The approach here — and I think this is the right one, but happy to hear other ideas — is to leave the dedup completely alone and fix the id instead: EventKit ids become `calendarItemIdentifier` + occurrence start date. True duplicates still collapse; occurrences survive. It also brings EventKit in line with Google, which already produces per-instance ids (`singleEvents=true`), so everything downstream keyed on `MBEvent.id` (dedup, notification plan identities, dismissals, auto-join records) was arguably designed for per-occurrence ids all along, and EventKit was the odd one out. The alternative would be keeping the shared id and teaching the three dedup sites to key on `(id, startDate)` — smaller on paper, but it leaves the scheduler's id-keyed lookups pairing plans with the wrong occurrence, so fixing the id at the source seemed like the better approach. If the shared-id model was intentional and I'm missing why, very open to reworking this.

**One behavior change worth calling out:** the `eventId` handed to `meetingStart` AppleScripts changes format for Apple Calendar events — it used to be the raw `calendarItemIdentifier`, now it carries a start-date suffix and differs per occurrence of a recurring meeting. Google events always behaved that way, but if anyone's scripts store or compare eventIds for Apple Calendar events, they'll see new ids after updating. If that's too rude a change for user scripts, an easy softening is passing the raw identifier to scripts specifically — say the word and I'll add it.

Other follow-on effects (went through every `MBEvent.id` consumer):
- Dismissing one occurrence of a recurring meeting now dismisses just that occurrence (the menu already renders them as separate rows).
- One-time after updating: dismissals/auto-join records persisted under old-format ids stop matching and get pruned as their events end; pending notification requests under old identifiers are reconciled away as stale.
- `ical://ekevent/…` open-in-calendar URLs still use the raw item identifier (pinned by a test).

The EKEvent → MBEvent mapping also moved out of the private fetch loop into `mbEvent(from:calendar:customRegexes:)` so it's testable without calendar access.

## Checklist
- [ ] Localized — n/a, no user-facing strings changed
- [ ] Added to changelog — n/a, fixes an unreleased regression (introduced during 5.0 development)

1. In Apple Calendar, create a daily recurring meeting. Set MeetingBar to show events for "today and tomorrow".
2. Without this change the menu shows only today's instance; with it, tomorrow's section lists tomorrow's occurrence too, and it participates in next-event selection and notifications.

Screenshoot 2026-07-04 at 20 11 58 2@2x

Screenshoot 2026-07-04 at 20 11 30@2x

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mapping of calendar events so recurring occurrences display separately instead of being merged, with more accurate duplicate handling.
  * Ensured calendar detail links and AppleScript event identifiers point to the correct underlying event.

* **New Features**
  * Added a dedicated event identifier (`scriptIdentifier`) for AppleScript usage.

* **Tests**
  * Expanded automated coverage for event ID semantics, URL/identifier construction, and AppleScript parameter expectations to prevent regressions.


leits:
Hi @jamtur01!
Thanks for digging into this! The approach makes sense to me.

One small change before merging: please keep passing the raw `calendarItemIdentifier` to `meetingStart` AppleScripts. The occurrence-specific ID looks right for internal use, but we shouldn’t break existing scripts.

jamtur01:
Done!

leits (approved):
Thanks @jamtur01!

PR #952, opened 2026-07-13
Fix icon size in status bar and update Discord icon

mavi3:
### Status
**READY**

### Description

The icon of the next or current meeting was always being resized to 16x16. This made the status bar ignore the height value set in the provider, resulting in stretched icons.

Now the the size is only forced to `MenuStyleConstants.iconSize` in the `.asset` case because the `getIconForMeetingService` method for `.meetingService` case already gives the height defined in provider.

In addition, the latest version of Discord icon was added and used in the provider. (it existed an icon before but it wasn't used)
(The stretch in the status bar was found while testing the new icon)

In master
| |  with the next meeting on Discord | with the next meeting on meet, and discord after |
|-|-|-|
| Image|image| image | 
| Result |Discord's new icon looks stretched on the vertical axis and Meets looks normal| In the Meet case it's barely noticeable, but it is stretched, whereas Discord looks normal.|

In latest release on a different computer with 8x8 example
| |  with the next meeting on something else| with the next meeting on 8x8 |
|-|-|-|
| Image|image |  image | 
| Result | 8x8 looks normal| 8x8 looks stretched|

#### After

| |  with the next meeting on Discord | with the next meeting on 8x8| 
|-|-|-|
| Image|  image |  image | 
| Result | Discord looks normal| 8x8 looks normal|

## Tests or validation commands you ran

- Added a new test to validate that the icon size is respected in status bar
- Tested with:
```bash
make build            
make test             
make test-logic   
make lint             
make validate-strings 
```

## How to Reproduce

- Create a calendar meeting with a link of a provider that defines a `height` value (for example 8x8 with is more noticeable) as the next meeting 
- Open MeetingBar and the icon will look stretched in vertical axis
- Notice that if you move the meeting after another one, it will stay stretched and it will only look normal after you restart the app


leits (approved):
Thanks @mavi3!

Issue #872, opened 2025-10-23
Bug:


jamtur01:
### macOS version

26.0.1

### MeetingBar version

4.11.6

### Installation source

manually

### Calendars provider

macOS Calendar app

### Bug description

Hi - sadly still seeing what looks like issue #716 with version 4.11.6 on Tahoe 26.0.1. It's only when I try to enable a google calendar - either directly through the API or through the Calendar app. I see some log entries that might be relevant:

```
11: 2025-10-23 10:07:41.748207-0400 0x54efcf5  Fault       0x7d895d9            61738  14   MeetingBar: (Security) [com.apple.runtime-issues:Security] __delegate_identifier__:Performance Diagnostics__:::____message__:This method should not be called on the main thread as it may lead to UI unresponsiveness.
...
13: 2025-10-23 10:07:41.748837-0400 0x54efcf5  Fault       0x7d895da            61738  14   MeetingBar: (Security) [com.apple.runtime-issues:Security] __delegate_identifier__:Performance Diagnostics__:::____message__:This method should not be called on the main thread as it may lead to UI unresponsiveness.
...
15: 2025-10-23 10:07:41.748841-0400 0x54efcf5  Fault       0x7d895db            61738  14   MeetingBar: (Security) [com.apple.runtime-issues:Security] __delegate_identifier__:Performance Diagnostics__:::____message__:This method should not be called on the main thread as it may lead to UI unresponsiveness.
```

and

```
687: 2025-10-23 10:09:17.851007-0400 0x54f1136  Activity    0x7d8c806            62913  0    MeetingBar: (EventKit) CADEventPredicate
688: 2025-10-23 10:09:17.856446-0400 0x54f1136  Default     0x0                  62913  0    MeetingBar: (libxpc.dylib) [com.apple.xpc:connection] [0x82c474500] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
689: 2025-10-23 10:09:18.060470-0400 0x54f1136  Default     0x0                  62913  0    MeetingBar: (libxpc.dylib) [com.apple.xpc:connection] [0x82b0ea940] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
[...]

Issue #721, opened 2024-05-03
Bug: Memory leak


davens:
### macOS version

14.1.1

### MeetingBar version

4.7.1

### Installation source

AppStore

### Calendars provider

Google Calendar

### Bug description

Meeting bar after a quit and restart uses about 20MB which is great.

Over days it starts growing in RAM usage, I've seen it hit gigabytes. I then need to quit and restart it.

This memory leak needs addressing.

### Reproduction steps

Use meeting bar

Have it active for multiple days or weeks. Watch memory usage in Activity monitor

### Expected behavior

Memory usage should not increase by orders of magnitude. It should realistically stay under 50MB given that it starts at 20MB.

Dentrax:
![Screenshot 2024-07-10 at 23 53 23](https://github.com/leits/MeetingBar/assets/16493751/16178abf-e379-4515-9c17-ac280b1edecf)

I can confirm, it uses 1.8GB!

`4.8.0`

/cc @leits

maxdeviant:
I'm seeing this as well on v4.8.1.

The memory usage starts off small and then grows over time:

<img width="905" alt="Screenshot 2024-07-28 at 8 20 47 AM" src="https://github.com/user-attachments/assets/c5a68e54-53d3-4559-9350-4f406ec2c80c">

Using it with Google Calendar with a single calendar.

exadeci:
Definitely a leak happening that being said my uptime is 104 days
<img width="827" alt="Screenshot 2024-09-26 at 9 25 57 AM" src="https://github.com/user-attachments/assets/8f779e02-5669-4c0b-af17-4f25d5dcde26">

davens:
Still happening on 4.10.0 @leits

leits:
Hi folks,
Thanks again for reporting this performance issue. I’ve just published a 4.11 beta release with a number of under-the-hood performance optimizations in MeetingBar. You can grab it here:
https://github.com/leits/MeetingBar/releases/tag/v4.11.beta 

Once you’ve had a chance to try it, could you please let me know whether you still see the issues? Any logs, profiled timings, or specific repro steps you can share will be super helpful. Thanks!

leits:
I have finalized the release and am gradually rolling it out, you can try it here:
https://github.com/leits/MeetingBar/releases/tag/v4.11.2

Issue #716, opened 2024-04-19
Bug: High CPU usage


rafaelzlisboa:
### macOS version

14.4.1 (23E224)

### MeetingBar version

4.8.0

### Installation source

Homebrew

### Calendars provider

Google Calendar

### Bug description

First of all, thank you very much for MB! Really appreciate the work you folks have put into the app, which is super useful for me and others on my team :)

I'm opening this issue because I'm experiencing high CPU usage on MeetingBar. On my work 16-inch, 2021 Apple M1 Pro it's consistently taking 10-20% CPU.

### Additional context

Seems similar to issues reported before #597 and #623. 

Had this problem on 4.6.0 and updated to 4.8.0 and the problem persists.

In search of anything more useful for you to work with, I went to the Console app and tried to find the MeetingBar application logs. 
Found it interesting that this line seems to appear a lot, though I couldn't quite figure what it means or if that's expected or if that's the issue (didn't find this exact line on the source code):
```
default	21:32:50.458985+0200	MeetingBar	[0x1166a7490] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
default	21:32:54.936252+0200	MeetingBar	[0x1267806a0] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
default	21:32:55.193105+0200	MeetingBar	[0x124e9f280] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
default	21:32:55.218806+0200	MeetingBar	[0x1064281b0] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
default	21:32:55.332218+0200	MeetingBar	[0x10653dac0] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
default	21:32:55.444941+0200	MeetingBar	[0x1246c7260] activating connection: mach=false listener=false peer=false name=com.apple.textkit.nsattributedstringagent
```

rafaelzlisboa:
Some new findings:

When I checked yesterday the CPU usage was normal (<1%). I pulled up the Console and recorded some logs and didn't see anything related to "activating connection". In fact, there seemed to be no logs at all from MeetingBar for the period it was running.
Now I checked again and it's back up to 10%-20% usage, and I see the `activating connection` logs.
I didn't reboot my computer over the weekend neither restart the application

KenSmz:
Hello,

[...]

PR #953, opened 2026-07-14
Strip HTML notes without leaking mach ports

jamtur01:
### Status
**READY**

### Description

Fixes a mach-port leak that gets MeetingBar force-killed after long uptime.

`htmlTagsStrippedForMeetingLinks` (link detection) and `String.htmlTagsStripped()` (notes display) both turned HTML notes into plain text with `NSAttributedString(html:)`. That initializer spins up TextKit's `com.apple.textkit.nsattributedstringagent` XPC service and leaks ~2 mach ports per call that are never reclaimed. Link detection runs it for every event with HTML notes on every calendar refresh, so the process port table grows unbounded until the kernel kills the app with `EXC_GUARD` — "allocating too many mach ports" (I caught one at 305,834 ports; unified-log shows the `nsattributedstringagent` connection flood right after each EventKit `CADEventPredicate`).

The HTML strip dates back to #144 and still exists in the 5.0 codebase, so the 4.11 rewrite didn't remove it.

Replaced both call sites with `HTMLPlainText`, a pure-Foundation strip (in `MeetingLinkDetector.swift` so the logic target and app share it): block-level tags become newlines, other tags are removed, and entities are decoded — all numeric forms (`&#123;`/`&#x7B;`) plus a named table covering the five XML names, common symbols, and the full Latin-1 accented-letter set (café, München). No XPC, nothing to leak. Named entities outside the table are left as written (documented in the code); numeric coverage is total.

Provenance:
- `d646f98` — replace `NSAttributedString(html:)` with `HTMLPlainText`; add the regression + behavioral tests.
- `a35af58` — flesh out the named-entity table (the first cut narrowed the old full decoding); also swap `mach_task_self_` → `task_self_trap()` in the port test so it builds under `-strict-concurrency=complete`.

Likely the cause behind #872 (same `nsattributedstringagent` signature), and probably #716 / #721 / #888.

No breaking changes to external interfaces. One behavior note: the notes-display path now decodes numeric + Latin-1/common named entities rather than the full HTML named set; anything rarer shows as written instead of decoded.

1. **Observe the XPC churn (before):** with `NSAttributedString(html:)` in the strip, add a calendar event whose notes contain HTML (Google/Outlook invites do) and let it refresh over time. `log stream --predicate 'process == "MeetingBar"'` shows repeated `activating connection … com.apple.textkit.nsattributedstringagent` right after each EventKit `CADEventPredicate`.
2. **Measure the leak (before):** sample the process mach-port count (`mach_port_names` on the task, or the harness in `testHTMLStripDoesNotLeakMachPorts`) across many strip calls — it climbs ~2 ports/call and never drops. Left running, the port table fills and the kernel kills the app with `EXC_GUARD` / port-space exhaustion (seen at 305,834 ports).
3. **The fix (after):** same scenario — no `nsattributedstringagent` connections in the log, and the mach-port count stays flat across calls.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved meeting link detection in HTML-formatted content, including Google Calendar redirect links.
  * Enhanced HTML-to-plain-text conversion by decoding common entities, removing script and style content, preserving block-level line breaks, and normalizing whitespace.
  * Reduced resource growth during repeated HTML processing.
  * Removed an unwanted trailing newline from cleaned notes.

* **Tests**
  * Added coverage for HTML conversion, redirect handling, entity decoding, line breaks, and resource cleanup.

### DIFF
--- a/MeetingBar/Extensions/String.swift
+++ b/MeetingBar/Extensions/String.swift
@@ -28,20 +28,16 @@ extension String {
     }
 
     /// Returns a version of the string with all HTML tags removed, if any.
+    ///
+    /// Uses the pure-Foundation `HTMLPlainText` strip rather than
+    /// `NSAttributedString(html:)`, which leaks mach ports via TextKit's
+    /// `nsattributedstringagent` XPC service (see `htmlTagsStrippedForMeetingLinks`).
     /// - Returns: The string without HTML tags.
     func htmlTagsStripped() -> String {
         if !containsHTML {
             return self
         }
-        return autoreleasepool {
-            guard let dataUTF16 = self.data(using: .utf16) else {
-                return self
-            }
-
-            let attributedString = NSAttributedString(html: dataUTF16, options: [.documentType: NSAttributedString.DocumentType.html], documentAttributes: nil)
-
-            return attributedString?.string ?? self
-        }
+        return HTMLPlainText.from(self)
     }
 
     func fileName() -> String {
--- a/MeetingBar/Meetings/MeetingLinkDetector.swift
+++ b/MeetingBar/Meetings/MeetingLinkDetector.swift
@@ -314,24 +314,40 @@ extension String {
 /// symbols, and the full Latin-1 accented-letter set (café, München, …). Named
 /// entities outside this table are left as written; numeric coverage is total.
 enum HTMLPlainText {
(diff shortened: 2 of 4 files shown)


jamtur01:
From what I can see, the test failure is an unrelated flaky test.

PR #955, opened 2026-07-19
Fix blank status bar item when the no-icon sentinel is hidden

WGlegola:
### Problem

`ensureStatusBarButtonIsVisible` only restores the fallback app icon when `button.image == nil`. A next event with no meeting service resolves to the `no_online_session` sentinel icon, which is set as a **non-nil** image with `imagePosition = .noImage` (drawn-suppressed). So with an empty title the guard never fires and the status item can render **completely blank**.

Reachable configuration: *hide meeting title* + *event-specific icon* + *hide time*, when the next event has no detected meeting link.

### Fix

Treat a suppressed image (`imagePosition == .noImage`) as absent in the visibility guard, so the app-icon fallback still applies. One-line change; no effect on the common case (a visible title keeps the guard's `attributedTitle.string.isEmpty` clause false, so the fallback still doesn't fire).

### Test

Adds `test_renderStatusBarUsesFallbackWhenSentinelIconIsHidden`: renders with `.meetingService(nil)` and an empty title, then asserts the item stays visible (`imagePosition == .imageLeft`, image is not the sentinel). All existing fallback/icon tests still pass.

Follow-up to the CodeRabbit note on #942.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the status bar icon could appear blank.
  * Added a fallback to display a visible icon when no active meeting service is available.
  * Preserved the existing layout and empty title behavior.

### DIFF
--- a/MeetingBar/UI/StatusBar/StatusBarItemController.swift
+++ b/MeetingBar/UI/StatusBar/StatusBarItemController.swift
@@ -246,7 +246,10 @@ final class StatusBarItemController {
     }
 
     private func ensureStatusBarButtonIsVisible(_ button: NSStatusBarButton) {
-        guard button.image == nil,
+        // A set-but-suppressed image (imagePosition == .noImage, e.g. the
+        // "no_online_session" sentinel) is not visible, so treat it the same as a
+        // missing image — otherwise the status item can render completely blank.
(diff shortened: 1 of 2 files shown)


leits (approved):
Thanks, @WGlegola!

PR #966, opened 2026-08-11
Fix Google Meet PWA launching

### LINKED ISSUE #963: 5.0.0-rc2: "Google Meet PWA" service-specific option opens a Chrome tab instead of the PWA
### Environment

- MeetingBar **5.0.0-rc2 (build 171)**, DMG build from the releases page
- macOS ProductVersion: 26.5 BuildVersion: 25F71
- Google Chrome 151.0.7922.76 (Official Build) (arm64) 
- Calendar source: Google Calendar
- Fresh install
- Google Meet PWA installed as a Chrome App (`~/Applications/Chrome Apps.localized/Google Meet.app`), app id `kjgfgldnnfoeklkmfkjfagphfepbbdan`, installed under the `Default` Chrome profile

### Summary

Setting Preferences → Service-specific opening → Google Meet → **Google Meet PWA** has no effect. Meet links open as an ordinary Chrome tab, silently, as if `defaultBrowser` were still in effect. No error and no toast.

The underlying Chrome app-mode invocation works fine on this machine, so this looks like the built-in `googleMeetPWA` path not reaching it.

### Failing configuration

With **Google Meet PWA** selected in the picker:

```
$ defaults read leits.MeetingBar | grep -iE 'browser|meet|pwa'
defaultBrowser = {... "name":"Google Chrome", "path":"/Applications/Google Chrome.app" ...}
providerBrowsers = {
    "Google Meet" = googleMeetPWA;
}
```

So the preference is being persisted correctly — the failure is downstream of that.

### Isolation

Run by hand, this opens the Meet PWA window reliably every time:

```bash
"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  --profile-directory="Default" \
  --app-id=kjgfgldnnfoeklkmfkjfagphfepbbdan \
  --app-launch-url-for-shortcuts-menu-item="https://meet.google.com/xxx-xxxx-xxx"
```

The Chrome side is healthy, the app id is correct, and the PWA is in the default profile — so the usual profile-mismatch explanation doesn't apply here.

I then tried a custom browser entry pointing at an executable shell script (`/Users/<user>/bin/meet-pwa`) wrapping that same command. MeetingBar rejects it:

> Oops! Unable to open the link in meet-pwa — Ensure you have meet-pwa installed, or open such links in a web browser from the preferences instead.

The script is `chmod +x` and runs correctly from a shell. Given the 4.11.2 notes advertise "you can now set any executable as a 'browser'," this may be a second, separate issue — or that field may only ever have accepted `.app` bundles in practice.

### Working workaround

Wrapping the same command in a minimal AppleScript application that registers itself as an `https` handler works, and MeetingBar drives it without complaint:

```applescript
on open location this_URL
	set chromeBin to "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
	do shell script quoted form of chromeBin & ¬
		" --profile-directory=Default" & ¬
		" --app-id=kjgfgldnnfoeklkmfkjfagphfepbbdan" & ¬
		" --app-launch-url-for-shortcuts-menu-item=" & quoted form of this_URL & ¬
		" > /dev/null 2>&1 &"
end open location
```

Exported as an app, with `CFBundleURLTypes` → `https` added to `Info.plist` and an ad-hoc re-sign, then selected as the custom browser for Google Meet:

```
$ defaults read leits.MeetingBar providerBrowsers
{
    "Google Meet" = "{...\"name\":\"MeetPWA\",\"path\":\"/Applications/MeetPWA.app\"}";
}
```

This confirms MeetingBar's URL hand-off machinery is fine. It's specifically the built-in PWA option that doesn't get there.

### Questions / likely causes

1. **Does the built-in option actually invoke Chrome with the app-mode flags?** Since the PWA here is in `Default` and the manual command works, the profile-mismatch theory is ruled out on this machine — which points at the flags not being passed at all, or the code path short-circuiting before it runs.
2. **Is the app id hard-coded or discovered** from the installed Chrome App shim? If discovered, worth checking that discovery path.
3. **Could the RC-2 `authuser` change interact?** If the URL passed to `--app-launch-url-for-shortcuts-menu-item` carries an `authuser` parameter, it'd be worth confirming Chrome still treats it as in-scope for the app rather than falling back to a tab.

### Did this exist in 4.11.x?

Not applicable — the Google Meet PWA option is new in 5.0. On 4.11.x I achieved the same result by routing Meet links through an external URL handler.

Happy to attach debug logs or test a patched build. Related prior requests: #834, #747, #313.

EhsanAzish80:
## Summary

- resolve the installed Google Meet app shim through Launch Services
- open Meet URLs directly with the PWA instead of spawning Chrome
- avoid sandbox-sensitive filesystem probing and preserve the PWA’s Chrome profile

## Testing

- full macOS test suite: 436 tests passed
- focused `ProviderOpeningPolicyTests` passed
- Release configuration built successfully
- installed the Google Meet PWA locally and verified the handoff opens a separate Google Meet app window rather than a normal Chrome tab

Closes #963

## Summary by CodeRabbit

* **Bug Fixes**
  * Google Meet meetings now open directly in the installed Google Meet app when available.
  * Improved detection of the installed Google Meet app.
  * If opening the app fails or it is unavailable, meetings now fall back to the selected browser.
  * Successful app launches no longer open an additional browser window or add unnecessary recent items.
  * Launch failures now correctly identify Google Meet as unavailable.

### DIFF
--- a/MeetingBar/Meetings/MeetingOpener.swift
+++ b/MeetingBar/Meetings/MeetingOpener.swift
@@ -316,37 +316,51 @@ enum GoogleMeetPWAInstallation {
 }
 
 enum GoogleMeetPWALauncher {
-    static func launch(_ plan: GoogleMeetPWAOpenPlan) -> Bool {
-        openApplication(plan)
-        return true
+    typealias Completion = @Sendable (Bool) -> Void
+    typealias ApplicationOpener = @Sendable (
+        GoogleMeetPWAOpenPlan,
+        @escaping @Sendable (NSRunningApplication?, Error?) -> Void
+    ) -> Void
+
+    static func launch(
+        _ plan: GoogleMeetPWAOpenPlan,
+        completion: @escaping Completion
+    ) {
+        launch(plan, applicationOpener: openApplication, completion: completion)
     }
 
     static func launch(
         _ plan: GoogleMeetPWAOpenPlan,
-        applicationOpener: (GoogleMeetPWAOpenPlan) throws -> Void
-    ) -> Bool {
-        do {
-            try applicationOpener(plan)
-            return true
-        } catch {
-            AppMessageCenter.shared.post(.browserUnavailable(name: "Google Meet"))
-            return false
+        applicationOpener: @escaping ApplicationOpener,
+        completion: @escaping Completion
+    ) {
+        applicationOpener(plan) { application, error in
+            let didLaunch = application != nil && error == nil
+            if !didLaunch {
+                AppMessageCenter.shared.post(.browserUnavailable(name: "Google Meet"))
+            }
+            completion(didLaunch)
         }
     }
 
-    private static func openApplication(_ plan: GoogleMeetPWAOpenPlan) {
-        let configuration = NSWorkspace.OpenConfiguration()
-        configuration.activates = true
-        configuration.addsToRecentItems = true
+    private static func openApplication(
+        _ plan: GoogleMeetPWAOpenPlan,
+        completion: @escaping @Sendable (NSRunningApplication?, Error?) -> Void
+    ) {
+        let configuration = openConfiguration()
         NSWorkspace.shared.open(
             [plan.meetingURL],
             withApplicationAt: plan.applicationURL,
-            configuration: configuration
-        ) { application, error in
-            if application == nil || error != nil {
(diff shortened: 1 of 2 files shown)


leits (approved):
Thanks, @EhsanAzish80!

Issue #836, opened 2025-06-10
Bug: Events disappearing and need to refresh the calendar or close and open the app


serginator:
### macOS version

15.5

### MeetingBar version

4.11.5

### Installation source

Homebrew

### Calendars provider

Google Calendar

### Bug description

Last week I started to notice my meetings were not at the menu bar, and I had to close the app and open it again, or open the settings, go to the Calendars and click on `Refresh`. This happens every few minutes, then it continues working for maybe one or two hours, then I have to do it again. I don't see a pattern.

### Reproduction steps

I'm not sure, I just have it synced with my Google calendar, and it's like it disconnects from time to time and I have to refresh it to show events again or to close and open it again.

### Expected behavior

Events should appear always

serginator:
Thanks! I'll try the workarounds and keep updating the app when new releases are out :)

Issue #832, opened 2025-06-02
Bug: Google auth requests every few hours


Techneaux:
### macOS version

15.5

### MeetingBar version

4.11.2

### Installation source

Homebrew

### Calendars provider

Google Calendar

### Bug description

Since upgrading to newest rewritten core MeetingBar, every few hours, I am requested to re-auth with my google account. This did not occur with the previous version of MeetingBar. 

The "MeetingBar already has access to these four services" makes me wonder if me having two google accounts from which MeetingBar can choose could be the actual root cause of the problem. Not sure though!

Thank you for MeetingBar - it is a wonderful tool.

<img width="1181" alt="Image" src="https://github.com/user-attachments/assets/44c98e82-0565-4299-b35e-0e68b587ec1d" />

<img width="1132" alt="Image" src="https://github.com/user-attachments/assets/db20f052-4d6d-4af4-b66e-8c1c79d8cd2d" />

<img width="753" alt="Image" src="https://github.com/user-attachments/assets/774a3edc-7ec1-4bbf-8c13-6332ba6d3e7e" />

leits:
Thanks for the report! Working on it.

leits:
Released in 4.11.3. Homebrew update is on the way:
https://github.com/Homebrew/homebrew-cask/pull/214860

Techneaux:
Unfortunately it appears I closed the issue too fast @leits. I received one auth request last night and then again this morning:

<img width="1143" alt="Image" src="https://github.com/user-attachments/assets/20758e53-ae92-4e42-a28a-57ba7798ef8a" />

<img width="1121" alt="Image" src="https://github.com/user-attachments/assets/70f0ee2b-2a1a-4339-baca-6bcb1ae937bf" />

<img width="1103" alt="Image" src="https://github.com/user-attachments/assets/ee3821d4-3c85-461b-8ff5-0d75fa3a9bc5" />

leits:
Thanks for your help! Prepared the new release with silent token refresh instead of explicit promting:
https://github.com/leits/MeetingBar/releases/tag/v4.11.5

kanru:
Could it be related to google auth? I changed to use apple calendar integration instead, but I was prompted for reauth after a day for apple too.

leits:
Please, check out the new version: https://github.com/leits/MeetingBar/releases/tag/v4.11.6

leits:
Issue seems fixed. Reopen if it happens again.

scottsb:
[...]

PR #974, opened 2026-09-04
Fix Google Calendar events silently not refreshing

### LINKED ISSUE #869: Bug: Meetingbar breaks every day - events not refreshing
### macOS version

15.7.1 (24G231)

### MeetingBar version

4.11.6

### Installation source

AppStore

### Calendars provider

Google Calendar

### Bug description

I believe that every day for the last few days, Meetingbar has failed to load the events, and effectively does nothing. When I open the Meetingbar dropdown, it was showing yesterday's events grayed out. I then tried Quick Actions -> Refresh Sources, and that fixed it. I'm hoping that fixed it for good, but I expect it will fail silently again tomorrow and I'll miss a few more meetings. Any ideas?

### Reproduction steps

_No response_

### Expected behavior

_No response_

### Screenshots/screen recordings

_No response_

### Additional context

_No response_

seahindeniz:
Fixes #869 (and the same root causes behind #832 / #836).

MeetingBar silently stops refreshing Google Calendar events — typically overnight or after a commute — and only recovers when the user presses Refresh, switches providers, or signs in again.

There is no single trigger. **Every failure mode on the Google path is unbounded: it either hangs forever or destroys the stored session.** Nothing times out, nothing retries with backoff, and nothing reacts to the network coming back. That's why the symptoms differ so much between reporters ("re-authenticate every day", "logged out after a commute", "Zscaler not up yet", "stale events until Refresh Sources").

### 1. A transient token-refresh failure destroys the saved session

`GCEventStore.validAccessToken` checked the token before the error. But AppAuth, on a **transient** (network) refresh failure, invokes the action with the *previous, already-expired* access token alongside the error — [`OIDAuthState.m`](https://github.com/openid/AppAuth-iOS/blob/master/Sources/AppAuthCore/OIDAuthState.m):

```objc
} else {
  if (error.domain == OIDOAuthTokenErrorDomain) { ... }   // grant rejected
  else { ...didEncounterTransientError... }               // network failure
}
...
actionToProcess.action(self.accessToken, self.idToken, error);
```

`-accessToken` only returns `nil` once `_authorizationError` is set, which a network error does not do. So the expired token came back and the error was dropped:

1. Expired token → Google returns **401**
2. `classify` → `.retryWithForcedTokenRefresh`
3. `validAccessToken(forceRefresh: true)` → still offline → **same expired token again**
4. Retry → **401** → `.clearAuthAndThrowAuthRequired`
5. `clearAuthState()` → `Keychain.delete` → **refresh token gone, full re-consent**

A laptop waking before Wi-Fi or a VPN is up was enough to wipe the OAuth grant.

### 2. Requests never failed — they hung for up to 7 days

`waitsForConnectivity = true` suppresses `timeoutIntervalForRequest`; only `timeoutIntervalForResource` applies, defaulting to **604800 s**. A fetch started while offline didn't error, it parked.

### 3. A background refresh could open an OAuth browser tab and await it forever

`ensureSignedIn()` → `signIn()` opened a browser and awaited a callback that only arrives if the user finishes. If they didn't, the fetch never completed — and `signInTask` was never cleared, so every later refresh awaited that dead task.

### 4. `flatMap(maxPublishers: .max(1))` turned one hung fetch into a total deadlock

Serializing fetches is only safe if the inner publisher always completes. Harness output for the exact pipeline shape:

```
=== healthy ===                       === first fetch hangs ===
[0.51s] fetch #1 STARTED              [0.50s] fetch #1 STARTED
[0.56s] fetch #1 DELIVERED            [2.00s] >>> user presses Refresh
...                                   RESULT: started=1 delivered=0
RESULT: started=9 delivered=8
```

A single never-completing fetch killed the 180 s timer, settings-change triggers **and the Refresh button** for the life of the process. `changeEventStoreProvider` is the only thing that cancels `refreshCycleTask` — which is exactly why "switch to Apple Calendar and back" is the workaround people keep finding.

### 5. No network-recovery trigger

Wake fires *before* Wi-Fi reassociates or a VPN connects — the worst possible moment given 1 and 2.

## Changes

- `validAccessToken` inspects `error` **before** `accessToken`; new retryable `AuthError.temporarilyUnavailable`. A forced refresh no longer reuses an in-flight ordinary refresh (which could only return the same rejected token).
- An API 401 surfaces `.throwAuthRequired` (visible Reconnect) instead of deleting the keychain entry. Session clearing stays with the token endpoint rejecting the refresh token, which AppAuth already reports via `didEncounterAuthorizationError`.
- `waitsForConnectivity = false` plus explicit 30 s / 90 s timeouts, so failures are bounded and the next cycle self-heals.
- Fetches use a non-interactive session gate; interactive authorization stays on the explicit sign-in entry point used by provider selection and Reconnect. `signIn` retries once with forced consent when the completed flow yielded no refresh token, preserving the recovery path `ensureSignedIn` used to provide.
- A timeout around the refresh cycle so the `Future` always completes and `.max(1)` cannot deadlock.
- An `NWPathMonitor` trigger on the unreachable → reachable transition, with a 60 s floor so a flapping VPN can't cause repeated full fetches.

## Testing

Unit tests updated for the renamed HTTP decision case in both suites, plus coverage for the transient-error description.

**Please note:** I developed this on a machine without Xcode (Command Line Tools only), so I could not run `make test` or build the app target locally. I type-checked the isolated logic file and the new concurrency shapes under `-swift-version 6 -strict-concurrency=complete`, and verified the deadlock fix with the harness above. **CI is the first real build of this branch** — happy to fix anything it turns up promptly.

One caveat on scope: none of this can eliminate a genuine Google-side revocation (user revokes access, or an OAuth client in "Testing" publishing status expiring refresh tokens after 7 days). That legitimately needs re-consent — but it now surfaces as an actionable Reconnect state rather than silence plus stale data.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01BruYnzM3jxEovNQYCdnhmd

## Summary by CodeRabbit

- **New Features**
  - Calendar data now refreshes automatically when the network becomes available again.
  - Google Calendar sign-in better handles temporary service interruptions while preserving valid sessions after authorization errors.

- **Bug Fixes**
  - Calendar refreshes now time out after two minutes instead of waiting indefinitely.
  - Google token refreshes now time out after 90 seconds, preventing stalled synchronization.
  - Reduced unnecessary interactive sign-in prompts during calendar and event fetching.
  - Improved handling of temporary authorization service interruptions without unnecessarily clearing stored sessions.

### DIFF
--- a/MeetingBar/App/AppDelegate.swift
+++ b/MeetingBar/App/AppDelegate.swift
@@ -183,6 +183,9 @@ class AppDelegate: NSObject, NSApplicationDelegate {
         lifecycleObserver.onDidWake = { [weak self] in
             self?.appModel?.handleWake()
         }
+        lifecycleObserver.onNetworkBecameReachable = { [weak self] in
+            self?.appModel?.handleNetworkBecameReachable()
+        }
         lifecycleObserver.onSystemClockChanged = { [weak self] in
             self?.handleSystemClockChange()
         }
--- a/MeetingBar/App/AppModel.swift
+++ b/MeetingBar/App/AppModel.swift
@@ -90,6 +90,7 @@ enum AppAction {
     case systemClockChanged
     case timezoneChanged
     case dayChanged
+    case networkBecameReachable
 
     // Calendar
     case calendarStoreChanged
@@ -331,7 +332,8 @@ final class AppModel: ObservableObject {
     func send(_ action: AppAction) {
         switch action {
         case .launched, .willTerminate, .screenLocked, .screenUnlocked,
-             .didWake, .systemClockChanged, .timezoneChanged, .dayChanged:
+             .didWake, .systemClockChanged, .timezoneChanged, .dayChanged,
+             .networkBecameReachable:
             handleLifecycleAction(action)
         case .calendarStoreChanged, .refreshCalendars, .calendarsLoaded,
              .eventsLoaded, .selectedCalendarsChanged, .providerHealthChanged,
@@ -363,6 +365,7 @@ final class AppModel: ObservableObject {
     func handleSystemClockChange() { send(.systemClockChanged) }
     func handleTimezoneChange() { send(.timezoneChanged) }
     func handleDayChange() { send(.dayChanged) }
+    func handleNetworkBecameReachable() { send(.networkBecameReachable) }
     func handleCalendarStoreChange() { send(.calendarStoreChanged) }
     func requestRefresh() { send(.refreshCalendars) }
     func reconcileNotifications() { send(.reconcileNotifications) }
@@ -425,7 +428,7 @@ final class AppModel: ObservableObject {
             state.timeContextRevision += 1
             reconcileNotificationsFromState()
             scheduleRefresh()
-        case .didWake, .dayChanged:
+        case .didWake, .dayChanged, .networkBecameReachable:
             scheduleRefresh()
         default:
             break
--- a/MeetingBar/App/LifecycleObserver.swift
+++ b/MeetingBar/App/LifecycleObserver.swift
@@ -36,6 +36,12 @@ final class LifecycleObserver {
     /// `nil` until the first path update, so becoming reachable at launch does
     /// not duplicate the refresh that launch already triggers.
     private var isNetworkReachable: Bool?
+    private var lastReachabilityCallback: Date?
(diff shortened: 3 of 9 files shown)


seahindeniz:
Hi @leits could you look at the issue and the PR, please 🙏🏻

